PR #1, opened 2024-01-16
Use Poetry to lock package versions

kongzii:


setup.py
kongzii, line 1: Removed setup.py as well, as it's not needed with the new project standard.

pyproject.toml
```diff
@@ -0,0 +1,27 @@
...
+pyautogen = "*"
+beautifulsoup4 = "*"
+serpapi = "*"
+google-search-results = "*"
+pytest = "*"
+llama-index = "~0.9.0"
+metagpt = "*"
+duckduckgo-search = "*"
+# crewai = "*"  # Conflicts with metagpt due to langchain, commented out as metagpt seems to have bigger traction, install manually in the environment for usage.
+
```
evangriffiths, line 24: I'm less familiar with poetry vs pip. Do you still get the local package installed with `poetry install`? Or do you need something like this here:
```
[tool.poetry.dev-dependencies]
prediction-market-agent = { path = ".", develop = true }
```
(added when you do `poetry add --dev .`) for that to work?

evangriffiths (changes requested):

prediction_market_agent/agents/crewai.py
```diff
@@ -7,56 +6,66 @@
 # TODO can use langchain's @tool decorator on our own tool methods to create a
 # tool useable by a crew agent
 
+try:
+    from crewai import Agent, Task, Crew
```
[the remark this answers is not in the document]
kongzii: They are also needed in the `run` method. Would you prefer to do import in every method of a class?

pyproject.toml
kongzii: Poetry won't allow to lock it with Crewai because of conflicts. However, once you have the poetry environment setup, you can do `pip install crewai` to force installation.

I'm not sure what's the best way to proceed here, as we are experimenting with a lot of experimental packages. If we find out we want CrewAi instead of Metagpt we can comment out Metagpt instead.

prediction_market_agent/agents/crewai.py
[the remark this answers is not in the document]
evangriffiths: We could import `Agent` in the `__init__` method inside a try-except block, and then inside `run` import `Task` and `Crew`, assuming the `crewai` package is installed.

pyproject.toml
evangriffiths: Ah sorry, I'm not referring to the commented out crewai package here. I just mean does doing `poetry install` also install the local `prediction_market_agent` package? If not then running `python main.py` wouldn't work, so I guess you've tested that!

prediction_market_agent/agents/crewai.py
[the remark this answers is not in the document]
kongzii: Sounds good, changed

pyproject.toml
kongzii: Aaa sorry! I somehow auto-assumed the comment was for the `crewai` and didn't notice you actually wrote `prediction-market-agent` in the code block.

Yes, it does, `python main.py` works as expected

evangriffiths: Ok sweet

evangriffiths (approved):

PR #2, opened 2024-01-16
Add pick_binary_market for Omen

kongzii:


prediction_market_agent/models.py
```diff
@@ -0,0 +1,10 @@
+from pydantic import BaseModel
+
+
+class Market(BaseModel):
```
kongzii, line 4: Moved to another file so we can reuse it easily across markets.

evangriffiths (approved):
Great!

PR #5, opened 2024-01-17
Integrate custom agent, without using any framework

kongzii:


prediction_market_agent/agents/custom_agent.py
```diff
@@ -0,0 +1,119 @@
...
+            if cycle_count == self.max_cycles - 1:
+                self.verbose_print(
+                    f"Agent: Reached {cycle_count} cycles, asking to provide answer now."
+                )
+                messages.append(
+                    "You have reached maximal message count, please provide answer now as best as you can."
+                )
+
+            # Because of the system prompt, completions are expected to be JSON-parseable.
+            completion_str = self.model.complete(messages)
```
kongzii, line 82: Interestingly, it works nicely without specifying other roles, just `system` and `user` here. But if we want to improve this model, the use of `assistant` and `tool` roles could help (at least in the case of OpenAI models).

prediction_market_agent/agents/langchain_agent.py
```diff
@@ -1,7 +1,7 @@
 from langchain.agents import load_tools
 from langchain.agents import initialize_agent
 from langchain.agents import AgentType
-from langchain.llms import OpenAI
+from langchain_community.llms import OpenAI
```
kongzii, line 4: Again based on Langchain's warning message

pyproject.toml
```diff
@@ -9,18 +9,20 @@ readme = "README.md"
...
 beautifulsoup4 = "*"
 serpapi = "*"
 google-search-results = "*"
 pytest = "*"
 llama-index = "~0.9.0"
-metagpt = "*"
 duckduckgo-search = "*"
-# crewai = "*"  # Conflicts with metagpt due to langchain, commented out as metagpt seems to have bigger traction, install manually in the environment for usage.
+crewai = "*"
+# metagpt = "*" # Commented out because requires super old version of langchain, and conflicts with crewai.
```
kongzii, line 25: Enabled Crewai and commented out Metagpt so we can use newer Langchain stuff.

prediction_market_agent/agents/custom_agent.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+
+        self.verbose = verbose
+        self.max_cycles = max_cycles
+        self.keys = utils.get_keys()
+        self.tools = {
+            "GoogleSearchTool": google_search,
+            "WebScrapingTool": tool_exception_handler(
+                map_exception_to_output={
+                    requests.exceptions.HTTPError: "Couldn't reach the URL. Don't scrap the same domain again."
```
evangriffiths, line 31:
```suggestion
                    requests.exceptions.HTTPError: "Couldn't reach the URL. Don't scrape the same domain again."
```

prediction_market_agent/agents/custom_agent.py
```diff
@@ -0,0 +1,119 @@
...
+
+        while True:
+            cycle_count += 1
+            # If we reached the maximum number of cycles, ask to provide an answer based on available information.
+            if cycle_count == self.max_cycles - 1:
+                self.verbose_print(
+                    f"Agent: Reached {cycle_count} cycles, asking to provide answer now."
+                )
+                messages.append(
+                    "You have reached maximal message count, please provide answer now as best as you can."
```
evangriffiths, line 78: I think this needs handling for the case where `cycle_count >= self.max_cycles`, which could happen if this isn't respected, and there isn't an `"answer"` key in the `completion_dict` in the following completion.

prediction_market_agent/agents/custom_agent.py
evangriffiths: Interesting. Maybe it's happening by default under the hood if not specified?

kongzii: Inside of `ChatOpenAIModel.complete`, I'm automatically assigning the `user` role to these messages, so I don't think they re-write it. But after all, it will all just be merged into one prompt, so maybe the model isn't super-sensitive about it.

prediction_market_agent/agents/custom_agent.py
```diff
@@ -0,0 +1,124 @@
...
+            else:
+                raise ValueError(f"Can not process the completion: {completion_str}")
+
+        assert answer is not None, "No answer was given."
+        return utils.parse_result_to_boolean(answer)
+
+
+if __name__ == "__main__":
+    agent = CustomAgent(verbose=True)
+    agent.run("Will the price of GNO be above $1000 at the end of the 2024?")
```
evangriffiths, line 124: Haha nice question :)

prediction_market_agent/agents/custom_agent.py
```diff
@@ -0,0 +1,124 @@
...
+            )(web_scrape_structured_and_summarized),
+        }
+        self.system_prompt = f"""You are a research assistant meant to answer questions placed on a predictive market.
+
+To do that, you can't rely on your training data, but you have access to the following tools in order to answer question:
+
+1. GoogleSearchTool: Given a query, returns a list of URLs. Accepts arguments as a dictionary, with a single key "query".
+2. WebScrapingTool: Given an objective and url, returns the summarized content of the page. In the objective, ask closed-ended questions. Accepts arguments as a dictionary, with keys "objective" and "url".
+
+If you want to use the tool, simply return completion in form of a dictionary with the keys "tool_name" and "tool_params".
```
evangriffiths, line 42: This actually works well?! I'm surprised you are getting good function args with this. Have you seen that you can pass in `tools` as an arg to `client.chat.completions.create`? (see https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools). Any reason you chose to stick them in the system prompt yourself instead of using that?

main.py
```diff
@@ -19,12 +20,13 @@ class Market(Enum):
 
 
 agent_mapping = {
-    AgentType.LANGCHAIN: pma.agents.langchain.LangChainAgent,
-    AgentType.AUTOGEN: pma.agents.autogen.AutoGenAgent,
+    AgentType.LANGCHAIN: pma.agents.langchain_agent.LangChainAgent,
```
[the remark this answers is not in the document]
evangriffiths: Hmm, I don't think this should be a problem. What's the issue you are seeing?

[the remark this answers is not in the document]
kongzii: For example, if I rename `autogen_agent.py` back to `autogen.py`, then this works

```
python main.py --agent-type autogen
```

But this fails

```
python prediction_market_agent/agents/custom_agent.py
```

with error

```
from autogen.agentchat.contrib.gpt_assistant_agent import GPTAssistantAgent
ModuleNotFoundError: No module named 'autogen.agentchat'; 'autogen' is not a package
```

prediction_market_agent/agents/custom_agent.py
kongzii: > This actually works well?!

Yeah, from my experience, GPT4 is quite good at following such requests.

> Any reason you chose to stick them in the system prompt yourself instead of using that?

I wanted to check out what will happen. If it will really work. Nice thing is that this is universal, we can try it out with other (open) models, too, without being constrained by OpenAI's format.

In the end the `tools` argument to OpenAI will convert it (somehow) to prompts too, but I'd expect from my implementation to have worse performance, due to different training.  But seems to be okay so far.

main.py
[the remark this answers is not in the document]
evangriffiths: Ah okay! That's annoying. I guess you could argue that we shouldn't be executing modules inside our package - that code should be in an `examples` or `tests` directory at the top level of the repo (so outside of the package), but that's fair enough!

prediction_market_agent/agents/custom_agent.py
evangriffiths: Makes sense. Yeah I guess performance depends a lot on how the system prompt compares with what's used in the training data. Cool that it works well!

PR #10, opened 2024-01-22
Add outcome selling function for Omen bets

kongzii:


scripts/bet_omen.py
```diff
@@ -26,14 +38,48 @@ def main(
...
         from_private_key=PrivateKey(from_private_key),
         market=market,
         outcome=outcome,
-        auto_deposit=True,
+        auto_withdraw=auto_withdraw,
     )
 
 
 if __name__ == "__main__":
-    typer.run(main)
```
kongzii, line 39: In case you didn't work with Typer before, it does pretty cool stuff out of the box: 
Screenshot by Dropbox Capture 
Screenshot by Dropbox Capture

evangriffiths: Oo I haven't used it before. Very nice!

prediction_market_agent/omen.py
```diff
@@ -275,9 +400,69 @@ def binary_omen_buy_outcome_tx(
...
+    approve_tx_receipt = omen_approve_all_market_maker_to_move_conditionaltokens_tx(
+        web3=web3,
+        market=market,
+        approve=True,
+        from_address=from_address,
+        from_private_key=from_private_key,
+        tx_params={"nonce": nonce},
+    )
+    nonce += ONE_NONCE  # Increase after each tx.
+    check_tx_receipt(approve_tx_receipt)
```
evangriffiths, line 440: Do you know what happens if you call `omen_sell_outcome_tx` on a market that has closed? Will this this check fail?

kongzii: It will not fail here, but it will fail while calling the contract, on something like `ContractLogicError: execution reverted: SafeMath: subtraction overflow`. I read that `receipt["status"]` can be zero (meaning there was some error), but so far all the errors I had were during the contract calling, not afterward from `check_tx_receipt`. But I add this anyway just to be sure.

evangriffiths (approved):

PR #12, opened 2024-01-23
Fix slippage in the calculate buy and sell functions

kongzii:


evangriffiths (commented):
Is this just moving the slippage calculation from the `calculate` functions to the level up, where they are called?

evangriffiths:
> Is this just moving the slippage calculation from the `calculate` functions to the level up, where they are called?

Ah, just seen your comment https://github.com/evangriffiths/prediction-market-agent/pull/11#discussion_r1463126800 !!

kongzii:
> Is this just moving the slippage calculation from the `calculate` functions to the level up, where they are called?

Yes, I think it makes more sense like this, anyway. From the names `omen_calculate_{buy,sell}_amount`, it's not clear they won't return the exact amounts.

evangriffiths (approved):
Just checked this commit with https://github.com/evangriffiths/prediction-market-agent/pull/11 and it fixes that TODO. Nice!!

PR #14, opened 2024-01-24
Fix omen manifold markets

kongzii:


agent.py
```diff
@@ -0,0 +1,63 @@
...
+
+import prediction_market_agent as pma
+from prediction_market_agent.markets.all_markets import MarketType, get_binary_markets
+from prediction_market_agent.agents.abstract import AbstractAgent
+from prediction_market_agent.agents.all_agents import AgentType, get_agent
+from prediction_market_agent.tools.utils import should_not_happen, check_not_none
+from prediction_market_agent.tools.types import Mana, xDai
+
+
+def main(
```
kongzii, line 14: There is a new agent.py entry point for infinite runs, and a bunch of TODOs are still here for following PRs .

main.py
```diff
@@ -1,78 +1,31 @@
...
 import prediction_market_agent as pma
-from prediction_market_agent.tools.types import xdai_type
+from decimal import Decimal
+from prediction_market_agent.agents.all_agents import AgentType, get_agent
+from prediction_market_agent.tools.types import xDai, Mana
 from prediction_market_agent.tools.utils import should_not_happen, check_not_none
+from prediction_market_agent.markets.all_markets import MarketType, get_binary_markets
 
 
-class AgentType(Enum):
```
kongzii, line 9: I just moved these Enums to different files, so it's nicely importable in the agent entry point, too.

main.py
```diff
@@ -1,78 +1,31 @@
...
-        "--auto-bet",
-        type=bool,
-        default=False,
-        help="If true, does not require user input to place the bet.",
-    )
-
-    parsed_args = args.parse_args()
-    selected_market = Market[parsed_args.market.upper()]
-    agent_type = AgentType[parsed_args.agent_type.upper()]
+def main(
```
kongzii, line 10: I changed to typer to simplify the code; hope you will like it! 

Screenshot by Dropbox Capture

prediction_market_agent/agents/abstract.py
```diff
@@ -1,8 +1,20 @@
...
+
+MarketVar = t.TypeVar("MarketVar", bound=MarketProtocol)
+
+
 class AbstractAgent:
     def __init__(self):
         pass
 
-    def run(self, objective: str) -> bool:
+    def pick_market(self, markets: list[MarketVar]) -> MarketVar:
```
kongzii, line 11: I added abstract pick_market for the infinite runs, but there is no implementation for this one yet.

prediction_market_agent/data_models/market_data_models.py
```diff
@@ -1,25 +1,54 @@
...
     HexAddress,
     ChecksumAddress,
+    Probability,
+    Mana,
+    OmenOutcomeToken,
 )
+from datetime import datetime
+
+
+class MarketProtocol(t.Protocol):
```
kongzii, line 16: I created this protocol to type functions that should work universally across all websites, with only the assumption that the market has string-representable ID and question.

agent.py
```diff
@@ -0,0 +1,63 @@
...
+
+    while True:
+        # TODO: Agent needs to keep track of the questions it has answered. It should skip them or re-evaluate.
+        available_markets = get_binary_markets(market_type)
+        logging.info(
+            f"Found {len(available_markets)} markets: {[m.question for m in available_markets]}"
+        )
+
+        # TODO: Pick the market with agent-specific logic, for now just pick the first one.
+        market = available_markets[0]  # agent.pick_market(available_markets)
```
evangriffiths, line 33: Since we're taking `available_markets[0]` here, `market` is an instance of the `Market` subclass, as apposed to `MarketProtocol`. So currently `MarketProtocol` is unused.

More generally though, it's not clear why we need separate `Market` and `MarketProtocol` classes. Why can't we hide all the different functionality of omen and manifold markets behind specialized methods on the `Market` base class?

prediction_market_agent/agents/abstract.py
kongzii: Yes! Updated.

prediction_market_agent/data_models/market_data_models.py
kongzii: It's a way to do "duck typing" or "structural typing" in Python. I'll try to explain on example:

Let's say we have protocol

```
class MyProtocol(t.Protocol):
  @property
  def question(self) -> str: ...
```

(it has to be written as @ property, I made a mistake here. I'm fixing that in the following types-fixing PR)

And two different classes + some function accepting the protocol

```
class A:
  question = "How are you?"

class B:
  question = "Do you like chihuahuas?"

def my_func(x: MyProtocol) -> None: ...
```

Now, both of these classes can be used in `my_func` and mypy won't complain, even though they are separate classes (they don't have a common parent)

```
my_func(A()) # OK
my_func(B()) # OK
```

are both correct, because they both implement the protocol (they have the `question` property).

---

Why I thought this is good for us? Because different markets (Omen, Manifold, ...) can have very different market structures. For example, Manifold has the `question` property, but Omen doesn't, so they cannot be subclasses of one common class.

But! We can easily add `question` property to Omen:

```
class OmenMarket(Market):
    ...

    @property
    def question(self) -> str:
        return self.title
```

And so agent's method can be typed as 

```
def answer_boolean_market(self, market: MarketProtocol) -> bool:
```

and now we know, that inside of `answer_boolean_market` we have object with `id` and `question` properties (hopefully all we need to make a decision), instead of doing something like `Union[OmenMarket, ManifoldMarket]`.

---

Hopefully my explanation make sense :-D

agent.py
```diff
@@ -0,0 +1,63 @@
...
+            from_address=check_not_none(keys.bet_from_address),
+            from_private_key=check_not_none(keys.bet_from_private_key),
+            market=market,
+            binary_outcome=answer,
+            auto_deposit=True,
+        ) if market_type == MarketType.OMEN else should_not_happen(
+            f"Unknown market: {market}"
+        )
+
+        logging.info(f"Sleeping for {sleep_time} seconds...")
```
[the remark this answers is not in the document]
kongzii: Nice, added!

prediction_market_agent/agents/abstract.py
```diff
@@ -1,8 +1,20 @@
...
+        """
+        Given a list of markets, pick one to answer.
+        """
+        raise NotImplementedError
+
+    def answer_boolean_market(self, market: MarketProtocol) -> bool:
```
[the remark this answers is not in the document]
kongzii: True, changed

agent.py
kongzii: > So currently MarketProtocol is unused.

It is used for type-checking.

> all the different functionality of omen and manifold markets behind specialized methods on the Market base class

How do you mean it here?

Something like

```
class Market(BaseModel):
  def get_question(self) -> str:
    return self.question if isinstance(self, ManifoldMarket) else self.title
```

? If so, then I'd say the Protocol gets rid of all the iffing for us. Results in a bit cleaner code and simpler typings.

prediction_market_agent/data_models/market_data_models.py
evangriffiths: Yes, lovely explanation, thanks!

I'm not sure if this is a taste thing, or if there's still something I'm missing though.

> Because different markets (Omen, Manifold, ...) can have very different market structures
Really isn't the issue just that the json response from querying the graph/manifold api gives a different key for the title/question?

To summarize, this solution here names the class properties based on the json keys, but then adds getters (`@property`) to reconcile the difference in naming:
```
"""
Option 1
"""
import typing as t
from pydantic import BaseModel

class MarketProtocol(t.Protocol):
    @property
    def question(self) -> str:
        ...

class Market(BaseModel):
    """Empty"""

class OmenMarket(Market):
    title: str

    @property
    def question(self) -> str:
        return self.title

class ManifoldMarket(Market):
    question: str

def print_question(m: MarketProtocol):
    print(m.question)

if __name__ == "__main__":
    print_question(OmenMarket.model_validate({"title": "What is the meaning of life?"}))
    print_question(
        ManifoldMarket.model_validate({"question": "What is the meaning of life?"})
    )
```

But equally we could solve this still by having a single baseclass that they derive from, and handle the name issue at construction time:
```
"""
Option 2
"""
from pydantic import BaseModel, Field

class Market(BaseModel):
    question: str  # Markets always have a 'question', regardless what the json calls it!

    def print_question(self):
        print(self.question)

class OmenMarket(Market):
    question: str = Field(..., alias="title") # From Omen subgraph

class ManifoldMarket(Market):
    """No specialization for now""" ""

if __name__ == "__main__":
    OmenMarket.model_validate(
        {"title": "What is the meaning of life?"}
    ).print_question()
    ManifoldMarket.model_validate(
        {"question": "What is the meaning of life?"}
    ).print_question()
```

Do you prefer Option 1 over Option 2?

agent.py
evangriffiths: Hopefully I've answered this in the other comment (https://github.com/evangriffiths/prediction-market-agent/pull/14#discussion_r1465173437)

prediction_market_agent/data_models/market_data_models.py
kongzii: I see what you mean now! 

> this solution here names the class properties based on the json keys

Yes, I thought this is the idea behind Pydantic, to have 1:1 mapping with JSON so we can see in the code what responses look like. But I agree that renaming can be handled with aliases.

> I'm not sure if this is a taste thing, or if there's still something I'm missing though.

One use-case for this is that it's only luck that Omen and Manifold have similar schemas. But they aren't the same market, what could happen is that we want to add a third one where structure is very different, for example:

```
{
  "id": "abc",
  "details": {
     "question": "How are you?"
   }
}
```

Now, we can't specialize `Market` to fit this schema, but what we can do is add a property returning `self.details.question`.

Another use-case is if the two markets don't return the same fields (in terms of their content, not names), but we know we need it for our agent (maybe to compute prediction cost vs. revenue or something). So we would have

```
class Market(BaseModel):
  ...
  usd_liquidity: USD
```

Here, we can't specialize Manifold's JSON because the information isn't there, but we can easily add a new property that will calculate it based on the `totalLiquidity: Mana` field. 

Screenshot by Dropbox Capture

evangriffiths: > I thought this is the idea behind Pydantic, to have 1:1 mapping with JSON

It doesn't have to be. Pydantic has the idea of 'validators' that allow you to map json to fields of a pydantic class, for cases like the one we have here (https://docs.pydantic.dev/latest/concepts/validators)

Doesn't it feel like we're introducing unnecessary complexity by matching the json schemas exactly? e.g. for this case the two examples are now:

(1)
```
import typing as t
from pydantic import BaseModel

class MarketProtocol(t.Protocol):
    @property
    def question(self) -> str:
        ...

class Market(BaseModel):
    """Empty"""

class OmenMarketDetails(BaseModel):
    title: str

class OmenMarket(Market):
    id: str
    details: OmenMarketDetails

    @property
    def question(self) -> str:
        return self.details.title

class ManifoldMarket(Market):
    id: str
    question: str

def print_question(m: MarketProtocol):
    print(m.question)

if __name__ == "__main__":
    print_question(
        OmenMarket.model_validate(
            {"id": "abc", "details": {"title": "What is the meaning of life?"}}
        )
    )
    print_question(
        ManifoldMarket.model_validate(
            {"id": "abc", "question": "What is the meaning of life?"}
        )
    )
```

(2)
```
from pydantic import BaseModel, model_validator

class Market(BaseModel):
    id: str
    question: str

    def print_question(self):
        print(self.question)

class OmenMarket(Market):
    @model_validator(mode="before")
    @classmethod
    def map_field_three(cls, data: dict) -> dict:
        if question := data.get("details", {}).get("title", None):
            data["question"] = question
        return data

class ManifoldMarket(Market):
    """No specialization for now""" ""

if __name__ == "__main__":
    OmenMarket.model_validate(
        {"id": "abc", "details": {"title": "What is the meaning of life?"}}
    ).print_question()
    ManifoldMarket.model_validate(
        {"id": "abc", "question": "What is the meaning of life?"}
    ).print_question()
```

By mapping the schemas to fit our abstraction of what a 'market' is reduces this complexity. I think the different markets are similar enough (i.e. our 'market' will be a clean enough abstraction) that this makes sense IMO.

> we can't specialize Manifold's JSON because the information isn't there, but we can easily add a new property that will calculate it based on the totalLiquidity: Mana field.

This is something we could also manage in `@model_validator` for `ManifoldMarket` - calculate the `usd_liquidity` at construction time.

What do you think of this approach?

kongzii: I thought using validators + not following their JSON schemas looks more complex in the end, vs. having one protocol.

I mean doing

```
@model_validator(mode="before")
    @classmethod
    def map_field_three(cls, data: dict) -> dict:
        if question := data.get("details", {}).get("title", None):
            data["question"] = question
```

vs.

```
@property
    def question(self) -> str:
        return self.details.title
```

but that's subjective for everyone, of course. 

---

```
    @model_validator(mode="before")
    @classmethod
    def map_field_three(cls, data: dict) -> dict:
        if question := data.get("details", {}).get("title", None):
            data["question"] = question
        return data
```

What I didn't like here is that we are losing the typing benefits. In this case, it's OK, as `question` will be None, and Pydantic will fail. 

But the whole `data` dictionary is untyped. Because we need `(mode="before")`, all of the Pydantic's type-checking isn't done yet.

And if we need to calculate USD from Mana, we will have to cast anything from the data dictionary into proper types + verify that it's ok as an input, and there is no IDE auto-completion.

Pydantic has also different behaviour with it, for example:

 ```
from pydantic import BaseModel, model_validator
from typing import Any


class Nested(BaseModel):
    field: str = "field"


class Parent(BaseModel):
    nested: Nested

    @model_validator(mode="before")
    @classmethod
    def map_field_three(cls, data: dict[str, Any]) -> dict[str, Any]:
        print(type(data["nested"]))
        return data


Parent(nested=Nested())
Parent.model_validate({"nested": {"field": "field"}})
```

will print 

```
peter/add-ci* $ python test.py                                                                                                                                        [15:13:33]
<class '__main__.Nested'>
<class 'dict'>
```

so there will be cases where we need to do `isinstance` magic for it to work.

---

I see in this id + question it looks fine, but I'm worried if there will be something more complex, it will become cumbersome.

---

What do you think about yet another option:

```

class OmenMarket(BaseModel):
  ...

class ManifoldMarket(BaseModel):
  ...

class AgentMarket(BaseModel):
  ...
  
  @staticmethod
  def from_manifold_market(market: ManifoldMarket) -> AgentMarket:
    ...

  @staticmethod
  def from_omen_market(market: OmenMarket) -> AgentMarket:
    ...
``` 

This way, there is no protocol weirdness, it's very similar to using `model_validator`, but it comes with all the type-checking. Also, it will be (probably) a lot simpler for Mypy, maybe we get rid of some bugs.

Comment (evangriffiths) [reply prediction_market_agent/data_models/market_data_models.py:17]: Haha, that's so diplomatic of you. Well I don't feel that strongly about (3) over (1). I suppose my objection was mainly over lack of familiarity with that pattern, which isn't a good reason to object.

One of the things that I was thinking you'd get from (2) as well is the ability to move the methods from prediction_market_agent/markets/all_markets.py to class methods for `OmenMarket` and `ManifoldMarket`. The reasoning being, I don't think you should have to add to this `if-else` list for each function in all_markets.py if you wanted to add a new market. But it doesn't make sense to call `market.place_bet(...)` either (it's the agent that places the bet on a market, so that's not a good abstraction).

So yeah, I would pick 3. But maybe we want something else for the above issue, but we can discuss that separately from this PR.

Comment (kongzii) [reply prediction_market_agent/data_models/market_data_models.py]: > market.place_bet(...)

Actually, I like this idea! It would get rid of the iff-ing, and I don't see anything bad coming out of it. We could do that in the next PR.

Implemented the  third method, check it out 🙏 Just instead of `from_omen_market` I did `to_agent_market` methods, so we don't need to do the iff-ing 🤞

File: prediction_market_agent/data_models/market_data_models.py [hunk 9]
```diff
@@ -1,16 +1,38 @@
...
+class AgentMarket(BaseModel):
+    """
+    Common market class that can be created from vendor specific markets.
+    Contains everything that is needed for an agent to make a prediction.
+    """
+
+    id: str
+    question: str
+    outcomes: list[str]
+    original_market: t.Union["OmenMarket", "ManifoldMarket"]
```
evangriffiths, line 26: Don't think we need this

evangriffiths (approved):
Looks great!! Thanks for being patient with my reviews

kongzii:
> Looks great!! Thanks for being patient with my reviews

Thanks for the feedback! <3

prediction_market_agent/data_models/market_data_models.py
kongzii: ( marking discussion moved to Slack so it doesn't look like I'm ignoring the comment 😄 )

PR #16, opened 2024-01-24
Fix typing

kongzii:


agent.py
```diff
@@ -32,7 +36,10 @@ def main(
             f"Found {len(available_markets)} markets: {[m.question for m in available_markets]}"
         )
 
-        market = agent.pick_market(available_markets)
+        market = t.cast(
+            t.Union[manifold.ManifoldMarket, omen.OmenMarket],
+            agent.pick_market(available_markets),
+        )  # TODO: Mypy bug: Works in VSCode type-checking, but doesn't in Mypy.
```
evangriffiths, line 42: Does
```
market: t.Union[manifold.ManifoldMarket, omen.OmenMarket] = agent.pick_market(available_markets)
```
not work?

prediction_market_agent/tools/google_search.py
```diff
@@ -4,7 +4,7 @@
 from prediction_market_agent import utils
 
 
-def google_search(query: str) -> t.List[str]:
```
evangriffiths, line 7: does mypy not like `t.List` or `t.Dict`?

prediction_market_agent/agents/abstract.py
```diff
@@ -5,10 +5,7 @@
 
 
 class AbstractAgent:
-    def __init__(self):
-        pass
-
-    def pick_market(self, markets: list[MarketVar]) -> MarketVar:
+    def pick_market(self, markets: t.Sequence[MarketVar]) -> MarketVar:
```
evangriffiths, line 8: why this change?

prediction_market_agent/markets/omen.py
```diff
@@ -440,11 +443,11 @@ def omen_sell_outcome_tx(
         web3=web3,
         market=market,
         approve=True,
-        from_address=from_address,
+        from_address=from_address_checksummed,
```
evangriffiths, line 446: bugfix?

prediction_market_agent/tools/betting_strategies.py
```diff
@@ -171,49 +184,53 @@ def _get_kelly_criterion_bet(
...
     max_bet: xDai,
 ) -> t.Tuple[xDai, OutcomeIndex]:
-    market: Market = get_market(market_address)
     if len(market.outcomeTokenAmounts) != 2:
         raise ValueError("Only binary markets are supported.")
 
-    # For FPMMs, the probability is equal to the marginal price
-    current_p_yes = Probability(market.outcomeTokenMarginalPrices[0])
-    outcome_index = 0 if estimated_p_yes > current_p_yes else 1
+    current_p_yes = market.outcomeTokenProbabilities[0]
```
evangriffiths, line 194: ah nice

evangriffiths (approved):
Mammouth effort!

agent.py
kongzii: It doesn't :( I'm not sure what's Mypy's problem here, with Pylance (Pyright) it's just okay: 

Screenshot by Dropbox Capture

prediction_market_agent/tools/google_search.py
kongzii: Sorry! It's my old habit to switch `t.X` to just `x[]`.

Do you prefer `t.List[..]`? Doing `list[...], set[...], dict[...]` is available from Python 3.9 without the need to import `typing`.

prediction_market_agent/agents/abstract.py
kongzii: Just to make Mypy happy. The sequence is immutable, but I think using `list` should be OK in this case. But Mypy doesn't believe so 😄

prediction_market_agent/markets/omen.py
kongzii: Only type-fix, code worked OK. But `get_transaction_count` used somewhere along the way requires a checksummed-type address.

prediction_market_agent/tools/google_search.py
evangriffiths: I guess I've been doing e.g. `t.List` instead of `list`, but I don't have a strong preference - we can just pick one. I'm happy to go with python types!

agent.py
kongzii: Not needed anymore, thanks to the removal of Protocol! 🎉

PR #23, opened 2024-02-06
Pass in generic BetAmount to `place_bet`

evangriffiths:
I prefer `place_bet` taking a single required `amount` arg, and then handling the conversion to the correct currency, as apposed to taking optional `amount`s in all the possible currencies it could support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,12 +37,16 @@ Execute `main.py` with optional arguments:
 
 ```bash
 % python main.py --help
-usage: main.py [-h] [--agent-type {langchain,autogen,always_yes}] [--auto-bet AUTO_BET]
+Usage: main.py [OPTIONS]
 
-optional arguments:
-  -h, --help            show this help message and exit
-  --agent-type {langchain,autogen,always_yes}
-  --auto-bet AUTO_BET   If true, does not require user input to place the bet.
+  Picks one market and answers it, optionally placing a bet.
+
+Options:
+  --market-type [manifold|omen]   [default: MarketType.MANIFOLD]
+  --agent-type [langchain|autogen|always_yes|coin_flip|llamaindex|metagpt|crewai|custom_openai|custom_llama]
+                                  [default: AgentType.ALWAYS_YES]
+  --auto-bet / --no-auto-bet      [default: no-auto-bet]
+  --help                          Show this message and exit.
 ```
 
 ## Testing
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -37,12 +37,16 @@ Execute `main.py` with optional arguments:
...
-  --agent-type {langchain,autogen,always_yes}
-  --auto-bet AUTO_BET   If true, does not require user input to place the bet.
+  Picks one market and answers it, optionally placing a bet.
+
+Options:
+  --market-type [manifold|omen]   [default: MarketType.MANIFOLD]
+  --agent-type [langchain|autogen|always_yes|coin_flip|llamaindex|metagpt|crewai|custom_openai|custom_llama]
+                                  [default: AgentType.ALWAYS_YES]
+  --auto-bet / --no-auto-bet      [default: no-auto-bet]
+  --help                          Show this message and exit.
```
evangriffiths, line 49: This is just an update since now `main.py` uses `typer`

kongzii: Thanks!

prediction_market_agent/data_models/market_data_models.py
```diff
@@ -32,7 +44,7 @@ class OmenMarket(BaseModel):
     https://aiomen.eth.limo
     """
 
-    BET_AMOUNT_CURRENCY: t.ClassVar[str] = "xDai"
+    BET_AMOUNT_CURRENCY: Currency = Currency.xDai
```
kongzii, line 47: This can be kept as a class var, or the field can have lower case name. Right now, it seems like a constant, but it's changeable when creating the class.

prediction_market_agent/markets/all_markets.py
```diff
@@ -10,6 +14,15 @@ class MarketType(str, Enum):
     OMEN = "omen"
 
 
+def get_bet_amount(amount: Decimal, market_type: MarketType) -> BetAmount:
+    if market_type == MarketType.OMEN:
+        return BetAmount(amount=amount, currency=Currency.xDai)
+    elif market_type == MarketType.MANIFOLD:
+        return BetAmount(amount=amount, currency=Currency.Mana)
+    else:
+        raise ValueError(f"Unknown market type: {market_type}")
```
kongzii, line 23: Simple and cool! Nicer than having N arguments for each currency :-D

kongzii (approved):

prediction_market_agent/data_models/market_data_models.py
evangriffiths: Ah good point!

PR #25, opened 2024-02-08
Fix auth issue with function scheduler

evangriffiths:
Using a http trigger for the cloud function causes failures when the scheduler tries to trigger the job. From what I could see this could be fixed by:

(1) switching to a pub/sub topic trigger for the function
(2) adding authentication to the scheduler

Couldn't see how to do (2) from the docs/googling, so have gone with (1)!


kongzii:
LGTM, we just need it to work for now 😄

kongzii (approved):

PR #41, opened 2024-03-26
Added LLM call as a way to save Tavily credits

gabrielfior:


prediction_market_agent/agents/known_outcome_agent/deploy.py
```diff
@@ -37,6 +38,12 @@ def pick_markets(self, markets: list[AgentMarket]) -> list[AgentMarket]:
             # been correctly bet on, and therefore the value of betting on them
             # is low.
             if not market_is_saturated(market=market):
+                if not has_question_event_happened_in_the_past(
+                    model=self.model, question=market.question
+                ):
+                    # event happened
```
kongzii, line 44: Wait, event happened but we are skipping it?

prediction_market_agent/agents/known_outcome_agent/deploy.py
```diff
@@ -37,6 +38,12 @@ def pick_markets(self, markets: list[AgentMarket]) -> list[AgentMarket]:
...
+                    continue
```
kongzii, line 45: This is level 6 of indentation, which is quite error-prone for the future.

Why not just 

```python
if not market_is_saturated(market=market) and has_question_event_happened_in_the_past(
    model=self.model, question=market.question
):
    ...
```

?

prediction_market_agent/agents/known_outcome_agent/known_outcome_agent.py
```diff
@@ -42,6 +43,22 @@ def has_known_outcome(self) -> bool:
...
+The current date is {date_str}. Your goal is to assert if a QUESTION references an event that is already finished (according to the current date and time) or if it will still take place in a later date. 
+
+For example, you should return 1 if given the event "Will Bitcoin have reached the price of $100 by 30 March 2023?", since the event ends on a data prior to the current date.
+
+Your answer MUST be an integer and follow the logic below:
+- If the event is already finished, return 1
+- If the event has not yet finished, return 0
+- If you are not sure, return -1
+
+Answer with the single prompt only, and nothing else.
```
kongzii, line 56:
```suggestion
Answer with the single 1, 0 or -1 only, and nothing else.
```

Just a terminology alert: `prompt` is what goes into the model, so the answer isn't a prompt, that's a completion.

gabrielfior: Thanks, adjusted.

prediction_market_agent/agents/known_outcome_agent/deploy.py
gabrielfior: That's misleading, adjusted thanks

kongzii (approved):

PR #46, opened 2024-04-03
Make KnownOutcomeAgent cheaper for unknown outcomes

evangriffiths:
TL;DR Before, markets without a known answer cost ~$0.40 in LLM credits, now they cost $0.05.

KnownOutcomeAgent predictions currently cost ~$0.03 for markets with known outcomes, but ~$0.40 for markets with unknown outcomes(!!). This makes them too expensive to run on a large number of markets.

Currently we work around this by disallowing the agent from betting on markets concerning events happening in the future (via `has_question_event_happened_in_the_past` for filtering markets). This is overly restrictive in that it filters out most of the Omen markets we care about betting on!
```
>>> q = "Will the Francis Scott Key Bridge in Baltimore be operational again by April 4th, 2024?"
>>> has_question_event_happened_in_the_past(model="gpt-4-1106-preview", question=q)
False
```

This PR removes this constraint and instead makes it a lot cheaper to research markets whose answer is not knowable, by allowing the agent to decide if a question is unknowable early, without having to exhaustively search and invoke the LLM until max retries is reached.

## Summary by CodeRabbit

- **New Features**
    - Expanded the `Result` options to better handle scenarios with unknowable outcomes.
- **Enhancements**
    - Improved market selection logic in prediction models, favoring markets with the earliest closing times or selecting randomly if no closing times are available.
- **Refactor**
    - Renamed a method for clarity in determining known results.
- **Style**
    - Updated prompt messages to align with the new outcome categorization.

### DIFF
--- a/prediction_market_agent/agents/known_outcome_agent/benchmark.py
+++ b/prediction_market_agent/agents/known_outcome_agent/benchmark.py
@@ -32,7 +32,9 @@ class QuestionWithKnownOutcome(BaseModel):
             id=self.question,
             question=self.question,
             p_yes=Probability(
-                self.result.to_p_yes() if self.result != Result.UNKNOWN else 0.5
+                self.result.to_p_yes()
+                if self.result != Result.KNOWN_UNKNOWABLE
+                else 0.5
             ),
             volume=None,
             created_time=None,
@@ -60,7 +62,8 @@ class KnownOutcomeAgent(AbstractBenchmarkedAgent):
             question=market_question,
             max_tries=self.max_tries,
         )
-        if answer.result == Result.UNKNOWN:
+        print(f"Answered {market_question=} with {answer.result=}, {answer.reasoning=}")
+        if not answer.has_known_result():
             return Prediction(
                 is_predictable=False,
                 outcome_prediction=None,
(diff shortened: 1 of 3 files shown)


kongzii:
```
>>> q = "Will the Francis Scott Key Bridge in Baltimore be operational again by April 4th, 2024?"
>>> has_question_event_happened_in_the_past(model="gpt-4-1106-preview", question=q)
False
```

I wonder why this doesn't work, I expected this kind of questions are super easy to evaluate 🤔

kongzii:
> I wonder why this doesn't work, I expected this kind of questions are super easy to evaluate 🤔

I know it now, because April 4th didn't happen yet! 😄 

Overall I like the idea of giving the agent option to quit early, for sure it's great, but do you think it would make sense to also improve the prompt of `has_question_event_happened_in_the_past`, if it's so errorous?

Feels like we should be able to tell that `... by <date X>` is in the past or not, if we can't do that, how good can our predictions be? (and for the questions without the exact date, we can continue and quit early thanks to these changes you made)

kongzii:
> again by April 4th, 2024?"

I guess the person who created that questions meant "by the end of April 3th" in reality, if that's often the case, we could change current date to `current date + 1 day` in the prompt for `has_question_event_happened_in_the_past`

kongzii (approved):
Withdrawing my comments after clarification on Slack, looks mega cool!

PR #51, opened 2024-04-05
Bump PMAT version to 0.9.3, simplify `calculate_bet_amount` for KnownOutcomeAgent

evangriffiths:
## Summary by CodeRabbit

- **Refactor**
	- Simplified the bet amount calculation logic for OmenAgentMarket.


prediction_market_agent/agents/known_outcome_agent/deploy.py
```diff
@@ -83,17 +82,7 @@ def answer_binary_market(self, market: AgentMarket) -> bool | None:
...
-            return BetAmount(
-                # On markets without liquidity, bet just a small amount for benchmarking.
-                amount=(
-                    Decimal(1.0)
-                    if market.get_liquidity_in_xdai() > 5
-                    else market.get_tiny_bet_amount().amount
-                ),
-                currency=Currency.xDai,
-            )
+            return BetAmount(amount=(Decimal(1.0)), currency=market.currency)
```
kongzii, line 85: I know we now filtering for markets with liquidity, but people can still create markets with for example, $0.01 in liquidity and then it doesn't make much sense to me to bet more than the minimal bet. wdyt?

kongzii: Don't you want to keep them for benchmark purposes? 

If not, okay by me, just that

```
elif market.get_liquidity_in_xdai() > 5:
                print(
                    f"Skipping market {market.id=} {market.question=}, because it has insufficient liquidity."
                )
```

Should it be with the opposite sign, probably?

evangriffiths: Oops! 😆

kongzii (approved):

prediction_market_agent/agents/known_outcome_agent/deploy.py
```diff
@@ -28,14 +27,19 @@ def market_is_saturated(market: AgentMarket) -> bool:
 
 
 class DeployableKnownOutcomeAgent(DeployableAgent):
-    model = "gpt-4-1106-preview"
+    model = "gpt-4-turbo-preview"
```
kongzii, line 30: I just noticed this as well, this can point to a different model in future without notice, is that desirable?

kongzii: Screenshot by Dropbox Capture

prediction_market_agent/agents/known_outcome_agent/deploy.py
evangriffiths: > Don't you want to keep them for benchmark purposes?

I'm not too concerned about that, more about trying to get the markets that should be p_yes/no==0.95 to that point

prediction_market_agent/agents/known_outcome_agent/deploy.py
evangriffiths: Hmm good question. Maybe in general not, but I reckon for this case it's good - my reasoning being that otherwise over time the agent will get less and less 'up to date with current events' which I would assume is bad for making future predictions.

PR #54, opened 2024-04-08
Implement BuyYes and BuyNo functions with Omen API

evangriffiths:


prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -51,13 +54,7 @@ def example_args(self) -> list[str]:
...
-        markets: list[AgentMarket] = OmenAgentMarket.get_binary_markets(
-            filter_by=FilterBy.OPEN,
-            sort_by=SortBy.CLOSING_SOONEST,
-            limit=5,
-        )
-
+        markets: list[OmenAgentMarket] = get_omen_binary_markets()
         market_questions_and_prices = []
         for market in markets:
             market_questions_and_prices.append(market.question)
```
gabrielfior, line 60: Maybe just a style preference, but I would prefer having some kind of tuple structure, like [(market_price, market_question)] instead of a list containing different item types. What do you think?

kongzii: Agreed, is there a reason why we can't use some Pydantic model here?

This seems like it could work:

```python
from pydantic import BaseModel


class MicroMarket(BaseModel):
    question: str
    p_yes: float

    def __str__(self) -> str:
        return f"'{self.question}' with probability of yes: {self.p_yes:.2%}"


mm = MicroMarket(question="Will it rain tomorrow?", p_yes=0.7)

prompt = f"""
You have following markets:

- {mm}
"""

print(prompt)
```

prints:

```bash
> python test.py                                                                                                                             [9:44:12]

You have following markets:

- 'Will it rain tomorrow?' with probability of yes: 70.00%
```

prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -98,44 +95,49 @@ def __call__(self) -> float:
...
-                f"Your balance of {balance} $ is not large enough to spend {amount} $."
-            )
-
-        balance -= amount
-        return "Bought " + str(amount * 2) + " yes outcome token of: " + market
-
+        return ["Will Joe Biden get reelected in 2024?", 2.3]
+
+    def __call__(self, market: str, amount: float) -> str:
+        if self.outcome == "yes":
```
kongzii, line 112: Outcomes on Omen are `Yes` and `No`, how come it's lowercased here? We have also constants for it:

https://github.com/gnosis/prediction-market-agent-tooling/blob/0e2adab90caf3aa0b73b5e9d126ef81465c5c43e/prediction_market_agent_tooling/markets/omen/data_models.py#L28

kongzii:
I see it currently uses Omen-specific classes/functions a lot, are you planning on having this deployable against Manifold as well? As we do with others?

prediction_market_agent/agents/microchain_agent/functions.py
evangriffiths: Nice, will do this!

I think this is still a starting point though, and we might want to do some thinking/experimenting how we want the api to look like for the agent. We could expose a bunch more things (opening time, closing time, liquidity, ...), either as extra fields of this pydantic class, or as separate `Functions` the agent can call.

My hunch is that separate functions makes more sense for token efficiency, but need to check.

Also, we could expose just the market Id here, and have the agent pass that into the other getter functions, so that we don't need `get_omen_market_by_title` here https://github.com/gnosis/prediction-market-agent-tooling/pull/172/files.

evangriffiths:
> I see it currently uses Omen-specific classes/functions a lot, are you planning on having this deployable against Manifold as well?

Yes, would like to support all markets we support. Once we figure out the set of Functions the agent needs, we can make them members of the `AgentMarket` base class. Haven't done that yet, because that's still in flux.

I would say that there isn't the same incentive to use manifold here because it's free, as for this agent I think token cost >> betting cost

prediction_market_agent/agents/microchain_agent/functions.py
evangriffiths: Good point, fixed.

PR #55, opened 2024-04-08
Implemented functions for microchain agent

gabrielfior:
- GetWalletBalance
- GetUserPositions

Depends on #54 before being merged

### DIFF
--- a/prediction_market_agent/agents/microchain_agent/functions.py
+++ b/prediction_market_agent/agents/microchain_agent/functions.py
@@ -1,13 +1,20 @@
+import os
 import pprint
 import typing as t
 
 from microchain import Function
-from prediction_market_agent_tooling.markets.agent_market import (
-    AgentMarket,
-    FilterBy,
-    SortBy,
-)
+from prediction_market_agent_tooling.markets.data_models import BetAmount, Currency
 from prediction_market_agent_tooling.markets.omen.omen import OmenAgentMarket
+from prediction_market_agent_tooling.tools.web3_utils import private_key_to_public_key
+
+from prediction_market_agent.agents.microchain_agent.tools import (
+    get_omen_binary_market_from_question,
+    get_omen_binary_markets,
+    get_omen_market_token_balance,
+)
+
+PRIVATE_KEY = os.getenv("BET_FROM_PRIVATE_KEY")
+PUBLIC_KEY = private_key_to_public_key(PRIVATE_KEY)
 
 balance = 50
 outcomeTokens = {}
@@ -51,13 +58,7 @@ class GetMarkets(Function):
         return []
 
     def __call__(self) -> list[str]:
-        # Get the 5 markets that are closing soonest
-        markets: list[AgentMarket] = OmenAgentMarket.get_binary_markets(
-            filter_by=FilterBy.OPEN,
-            sort_by=SortBy.CLOSING_SOONEST,
-            limit=5,
-        )
-
+        markets: list[OmenAgentMarket] = get_omen_binary_markets()
         market_questions_and_prices = []
         for market in markets:
             market_questions_and_prices.append(market.question)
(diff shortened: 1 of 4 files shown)


prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -209,6 +223,37 @@ def __call__(self, summary: str) -> str:
...
 
 
+class GetWalletBalance(Function):
+    @property
+    def description(self) -> str:
+        return "Use this function to fetch the balance of a user, given in xDAI units."
+
+    @property
+    def example_args(self) -> list[str]:
+        return ["0x2DD9f5678484C1F59F97eD334725858b938B4102"]
```
evangriffiths, line 233: I don't the agent needs to be aware of its own wallet address, so I think this function can take no arguments, and just internally use the `PUBLIC_KEY` global variable.

prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -209,6 +223,37 @@ def __call__(self, summary: str) -> str:
...
+
+
+class GetUserPositions(Function):
+    @property
+    def description(self) -> str:
+        return "Use this function to fetch the markets where the user has previously bet."
+
+    @property
+    def example_args(self) -> list[str]:
+        return ["0x2DD9f5678484C1F59F97eD334725858b938B4102"]
```
evangriffiths, line 249: Same here

prediction_market_agent/agents/microchain_agent/tools.py
```diff
@@ -0,0 +1,53 @@
+from eth_typing import HexStr, HexAddress, ChecksumAddress
```
evangriffiths, line 1: just as a heads up, I moved this to microchain_agent/utils.py. I thought 'tools' might be confusing because 'tools/functions' have a loaded meaning in the context of agents (i.e. some code that is registered with the agent, and it can call).

tests/tools/microchain_agent/test_functions.py
```diff
@@ -0,0 +1,58 @@
+import pytest
```
evangriffiths, line 1: In https://github.com/gnosis/prediction-market-agent/pull/54 I added tests/agents/test_microchain.py, so I guess these should be added there.

tests/tools/microchain_agent/test_functions.py
```diff
@@ -0,0 +1,58 @@
...
+    expected_condition_ids = [
+        HexBytes("0x9c7711bee0902cc8e6838179058726a7ba769cc97d4d0ea47b31370d2d7a117b"),
+        HexBytes("0xe2bf80af2a936cdabeef4f511620a2eec46f1caf8e75eb5dc189372367a9154c"),
+        HexBytes("0x3f8153364001b26b983dd92191a084de8230f199b5ad0b045e9e1df61089b30d"),
+    ]
+    unique_condition_ids = sum([u.position.conditionIds for u in user_positions], [])
+    assert set(expected_condition_ids).issubset(unique_condition_ids)
+
+def test_balance_for_user_in_market() -> None:
+    user_address = '0x2DD9f5678484C1F59F97eD334725858b938B4102'
```
evangriffiths, line 44: Use `AGENT_0_ADDRESS` here.

prediction_market_agent/agents/microchain_agent/tools.py
```diff
@@ -0,0 +1,53 @@
...
+            return m
+    raise ValueError(f"Market '{market}' not found")
+
+
+def find_index_set_for_market_outcome(market: OmenAgentMarket, market_outcome: str):
+    try:
+        market_outcome_match = market.outcomes.index(market_outcome)
+        # Index_sets start at 1
+        return market_outcome_match + 1
+    except ValueError as e:
```
evangriffiths, line 40: Is this still WIP? Don't think there should be a try-except here 😛

prediction_market_agent/agents/microchain_agent/tools.py
```diff
@@ -0,0 +1,53 @@
...
+from prediction_market_agent_tooling.markets.omen.omen import OmenAgentMarket
+from prediction_market_agent_tooling.markets.omen.omen_contracts import OmenConditionalTokenContract, \
+    OmenFixedProductMarketMakerContract
+from prediction_market_agent_tooling.markets.omen.omen_subgraph_handler import OmenSubgraphHandler
+from web3 import Web3
+from web3.types import Wei
+
+
+def address_to_checksum_address(address: str) -> ChecksumAddress:
+    return ChecksumAddress(HexAddress(HexStr(address)))
```
kongzii, line 16: This will change the typing so mypy will be happy, but in order to actually checksum the address, it's needed to call

```
Web3.to_checksum_address(self.collateralTokenAddress)
```

(if typing is changed to checksum address, but the address isn't actual checksummed, transactions can fail)

kongzii: And `to_checksum_address` is typed correctly on the library, so no need fur the wrapper function

gabrielfior:
Remove `getMarketByTitle` and stick with `getMarketById`

prediction_market_agent/agents/microchain_agent/tools.py
gabrielfior: Removed

gabrielfior:
Closed this in favor of #59

PR #59, opened 2024-04-09
Implemented functions for microchain agent

gabrielfior:
- getWalletBalance
- get markets given user positions
- fetch individual token balances for a given market outcome

PR #60, opened 2024-04-09
Make microchain Functions (somewhat) market-type agnostic

evangriffiths:
Subclass microchain `Function` class with `MarketFunction` that takes `MarketType`, to differentiate behaviour of functions.

This way the abstractions in microchain_agent/functions.py don't have any market-specific code in them (i.e. no 'Omen' or 'Maninfold' imports from PMAT). There are few left over after the merge of https://github.com/gnosis/prediction-market-agent/pull/59, but the idea is there.

### DIFF
--- a/prediction_market_agent/agents/microchain_agent/functions.py
+++ b/prediction_market_agent/agents/microchain_agent/functions.py
@@ -1,13 +1,14 @@
 import pprint
 import typing as t
 from decimal import Decimal
+from typing import cast
 
 from eth_utils import to_checksum_address
 from microchain import Function
+from prediction_market_agent_tooling.markets.agent_market import AgentMarket
 from prediction_market_agent_tooling.markets.data_models import BetAmount, Currency
+from prediction_market_agent_tooling.markets.markets import MarketType
 from prediction_market_agent_tooling.markets.omen.data_models import (
-    OMEN_FALSE_OUTCOME,
-    OMEN_TRUE_OUTCOME,
     OmenUserPosition,
     get_boolean_outcome,
 )
@@ -20,9 +21,11 @@ from prediction_market_agent_tooling.tools.balances import get_balances
 from prediction_market_agent.agents.microchain_agent.utils import (
     MicroMarket,
     fetch_public_key_from_env,
-    get_omen_binary_market_from_question,
-    get_omen_binary_markets,
-    get_omen_market_token_balance,
+    get_binary_market_from_question,
+    get_binary_markets,
+    get_market_token_balance,
+    get_no_outcome,
+    get_yes_outcome,
 )
 
 balance = 50
@@ -57,7 +60,13 @@ class Product(Function):
         return a * b
 
 
-class GetMarkets(Function):
+class MarketFunction(Function):
+    def __init__(self, market_type: MarketType) -> None:
+        self.market_type = market_type
+        super().__init__()
+
+
+class GetMarkets(MarketFunction):
     @property
     def description(self) -> str:
         return "Use this function to get a list of predction markets and the current yes prices"
@@ -68,11 +77,12 @@ class GetMarkets(Function):
 
     def __call__(self) -> list[str]:
         return [
-            str(MicroMarket.from_agent_market(m)) for m in get_omen_binary_markets()
+            str(MicroMarket.from_agent_market(m))
+            for m in get_binary_markets(market_type=self.market_type)
         ]
 
 
-class GetPropabilityForQuestion(Function):
+class GetPropabilityForQuestion(MarketFunction):
     @property
     def description(self) -> str:
         return "Use this function to research the probability of an event occuring"
@@ -90,7 +100,7 @@ class GetPropabilityForQuestion(Function):
         return 0.0
 
 
-class GetBalance(Function):
+class GetBalance(MarketFunction):
     @property
     def description(self) -> str:
         return "Use this function to get your own balance in $"
@@ -105,11 +115,11 @@ class GetBalance(Function):
         return balance
 
 
-class BuyTokens(Function):
-    def __init__(self, outcome: str):
-        self.user_address = fetch_public_key_from_env()
+class BuyTokens(MarketFunction):
+    def __init__(self, market_type: MarketType, outcome: str):
         self.outcome = outcome
-        super().__init__()
+        self.user_address = fetch_public_key_from_env()
+        super().__init__(market_type=market_type)
 
     @property
     def description(self) -> str:
@@ -122,11 +132,16 @@ class BuyTokens(Function):
     def __call__(self, market: str, amount: float) -> str:
         outcome_bool = get_boolean_outcome(self.outcome)
 
-        market_obj: OmenAgentMarket = get_omen_binary_market_from_question(market)
+        market_obj: AgentMarket = get_binary_market_from_question(
+            market=market, market_type=self.market_type
+        )
+        market_obj = cast(
+            OmenAgentMarket, market_obj
+        )  # TODO fix with 0.10.0 PMAT release
         outcome_index = market_obj.get_outcome_index(self.outcome)
         market_index_set = outcome_index + 1
 
-        before_balance = get_omen_market_token_balance(
+        before_balance = get_market_token_balance(
             user_address=self.user_address,
             market_condition_id=market_obj.condition.id,
             market_index_set=market_index_set,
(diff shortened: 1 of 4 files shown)


prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -191,24 +210,6 @@ def __call__(self, market: str, amount: int) -> str:
         return "Sold " + str(amount) + " no outcome token of: " + market
 
 
-class BalanceToOutcomes(Function):
```
evangriffiths, line 194: Have decided we don't need, so removing for now

prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -57,7 +60,13 @@ def __call__(self, a: float, b: float) -> float:
         return a * b
 
 
-class GetMarkets(Function):
+class MarketFunction(Function):
+    def __init__(self, market_type: MarketType) -> None:
+        self.market_type = market_type
+        super().__init__()
```
evangriffiths, line 66: @kongzii Do you have any thoughts on whether `self.market_type` should be set based on env var vs programatically, as I have done in this PR?

@gabrielfior and I are split on it.

From what I see pros of having it being programatic:
- consistent with how we deploy/benchmark/monitor agents in PMAT
- easy to parametrize on it in pytest

Pros of env var:
- consistent with other config vars like wallet address/pkey

gabrielfior: Some additional thoughts:
- If we do have an env like `MARKET_TYPE=OMEN`, we could restrict the tools the agent has access to by having some kind of mapping (market_type -> list[Function]), or even fancier, we could decorate the functions that belong to Omen as @OmenFunction, so that we can fetch all functions decorated like this programatically and pass them to the agent, depending on the env variable.
- Happy with both approaches here.

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -1,17 +1,20 @@
...
 load_dotenv()
 
 engine = Engine()
 engine.register(Reasoning())
 engine.register(Stop())
-for function in ALL_FUNCTIONS:
+for function in MISC_FUNCTIONS:
     engine.register(function())
+for function in MARKET_FUNCTIONS:
+    engine.register(function(market_type=MarketType.OMEN))
```
gabrielfior, line 17: I was wondering if we accomplish this via a decorator, e.g. we mark all classes that are market functions with @MarketFunction and retrieve those afterwards.
I created a quick-and-dirty example here -> https://gist.github.com/gabrielfior/3d777af14080a121a0cefa2ee2f172e9
The current implementation is fine, but we might forget to keep those lists updated.

gabrielfior (approved):

prediction_market_agent/agents/microchain_agent/functions.py
kongzii: > If we do have an env like MARKET_TYPE=OMEN, we could restrict the tools the agent has access to by having some kind of mapping (market_type -> list[Function])

Why that wouldn't be possible with a programmatic approach?

>  Do you have any thoughts on whether self.market_type should be set based on env var vs programatically, as I have done in this PR?

I'm not sure you guys will like my point of view, as it brings the most work, but here it goes! 😄 

Personally, I would do kind of both, in a way as in the PMAT repository:

- Remove all the `load_dotenv` callings, because modifying state just by importing files isn't nice
- Create single config class as in PMAT [here](https://github.com/gnosis/prediction-market-agent-tooling/blob/513f85cc55da6347758be822c1e16f1c5fd6b5fe/prediction_market_agent_tooling/config.py#L18), so we have all the configuration at one place, because if `os.environ.get` is used too much across the code base, then it's hard to keep track of all the configs available
- Initialise and use that class directly where it make sense or somewhere beforehand and just pass the concrete input arguments

In this case, it makes sense to me that `MarketFunction` accepts  `market_type: MarketType` as the init argument instead of deciding about it quietly in the background (another benefit I can see is that in the future, there could be multiple markets available to agent and that wouldn't be possible if `MarketFunction` behaviour is baked into environment variables)

prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -122,11 +132,16 @@ def example_args(self) -> list[t.Union[str, float]]:
     def __call__(self, market: str, amount: float) -> str:
         outcome_bool = get_boolean_outcome(self.outcome)
 
-        market_obj: OmenAgentMarket = get_omen_binary_market_from_question(market)
+        market_obj: AgentMarket = get_binary_market_from_question(
+            market=market, market_type=self.market_type
+        )
+        market_obj = cast(
+            OmenAgentMarket, market_obj
+        )  # TODO fix with 0.10.0 PMAT release
```
kongzii, line 140: <3

prediction_market_agent/agents/microchain_agent/microchain_agent.py
kongzii: > https://gist.github.com/gabrielfior/3d777af14080a121a0cefa2ee2f172e9

That's an interesting trick!

No preference from my side but it's cool

tests/agents/test_microchain.py
```diff
@@ -33,39 +34,46 @@ def before_all_tests() -> t.Generator[None, None, None]:
     yield None
 
 
-def test_get_markets() -> None:
-    get_markets = GetMarkets()
+@pytest.mark.parametrize("market_type", [MarketType.OMEN])
```
kongzii, line 37: Is it correct that they are tested only for Omen?

kongzii (approved):

prediction_market_agent/agents/microchain_agent/microchain_agent.py
evangriffiths: Oo yes nice trick. I think we need to have market-specific functions inherit from `MarketFunction` anyway though. Feels a bit clunky to me registering them separately like this, or can this be combined?

```
@market_function
class GetBalance(MarketFunction)
    ...
```

I know you can get subclasses of a class with:
```
import inspect

def find_imported_classes_by_type(base_class):
    classes = []
    for name, obj in globals().items():
        if inspect.isclass(obj) and issubclass(obj, base_class) and obj is not base_class:
            classes.append(obj)
    return classes

MARKET_FUNCTIONS = find_imported_classes_by_type(MarketFunction)
```

but then you have the issue where 'abstract' subclasses (i.e. `BuyTokens` in this case) will be added to `MARKET_FUNCTIONS`.

tests/agents/test_microchain.py
evangriffiths: Good point. For this one it should work for all markets (except isn't working) for polymarket atm. Will investigate separately.

The others are currently Omen specific, but once we have a 1-1 mapping for a AgentMarket.some_function -> Function we can do `@pytest.mark.parametrize("market_type", list(MarketType))`

prediction_market_agent/agents/microchain_agent/functions.py
evangriffiths: > Create single config class as in PMAT [here](https://github.com/gnosis/prediction-market-agent-tooling/blob/513f85cc55da6347758be822c1e16f1c5fd6b5fe/prediction_market_agent_tooling/config.py#L18)

Nooo, I want to remove that! (https://github.com/gnosis/prediction-market-agent-tooling/issues/71) 😄. But actually for a particular agent I think it makes much more sense than in the PMAT repo, as the keys required are fixed and known

kongzii: > But actually for a particular agent I think it makes much more sense than in the PMAT repo, as the keys required are fixed and known

Yep, not a great choice for tooling repo, but for the agent it sounds good to me

kongzii: Anyway, no need to do that in this PR. TLDR is that I like programic `__init__(self, market_type: MarketType)` more here :-D Doing it inside of __init__ from env vars will be error prone.

evangriffiths: It's a good point though - I've imported APIKeys here to replace our manual .env loading. So if it is removed from PMAT it can live on in this repo 😄

kongzii: I just remembered that we already have something in PMA repo too: https://github.com/gnosis/prediction-market-agent/blob/46d0ddae565909d8dd617e6448cbf40a4dc8c4d9/prediction_market_agent/utils.py#L11

It subclasses PMAT's API keys, so the same stuff is available in PMA's APIKeys as well, but once they are removed in PMAT, should be easier to just add required stuff here.

evangriffiths: Ah nice, I've switched to that here https://github.com/gnosis/prediction-market-agent/pull/63

PR #63, opened 2024-04-10
Fix GetWalletBalance.__call__ args

evangriffiths:
Added test to ensure call args for all registered functions are correct.


prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -1,13 +1,9 @@
-import os
```
evangriffiths, line 1: also using APIKeys here as well (unrelated to the fix)

kongzii (approved):

PR #64, opened 2024-04-10
Coinflip agent deployment

kongzii:


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+ENV VIRTUAL_ENV=/app/.venv \
+    PATH="/app/.venv/bin:$PATH"
+
+WORKDIR /app
+
+COPY --from=builder ${VIRTUAL_ENV} ${VIRTUAL_ENV}
+
+COPY prediction_market_agent ./prediction_market_agent
+
+ENV PYTHONPATH=/app
```
evangriffiths, line 29: Do we need a `CMD`/`ENTRYPOINT` here or is that done in the Kubernetes config? I guess if we're building this in CI then we want to test running it too?

kongzii: Good question, command is easily changeable and I put it into Terraform https://github.com/gnosis/gnosisai-infrastructure/pull/3/files#diff-fa25fb4ef6fa67ec2382f07d26b9039099ce7a314a7e98517370a8439a0d0365R42

But if I put it here, it's easier to change it without the need for updating Terraform 🤔 I'll give it a try!

kongzii: Yep works fine, updated 🙏

prediction_market_agent/agents/coinflip_agent/coinflip_agent.py
```diff
@@ -0,0 +1,13 @@
...
+from prediction_market_agent_tooling.deploy.agent import DeployableAgent
+from prediction_market_agent_tooling.markets.agent_market import AgentMarket
+
+
+class DeployableCoinFlipAgent(DeployableAgent):
+    def pick_markets(self, markets: t.Sequence[AgentMarket]) -> t.Sequence[AgentMarket]:
+        return random.sample(markets, 1)
+
+    def answer_binary_market(self, market: AgentMarket) -> bool | None:
+        return random.choice([True, False])
```
evangriffiths, line 13: Should we make it a convention to have:

1. prediction_market_agent/agents/<agent_name>/<agent_name>.py: core code
2. prediction_market_agent/agents/<agent_name>/deploy.py: subclassing DeployableAgent

?

I guess that would mean just renaming this file delploy.py

evangriffiths (approved):

prediction_market_agent/agents/coinflip_agent/coinflip_agent.py
kongzii: I see known_outcome_agent already uses (2), so I changed it to that one. Didn't notice it before, sorry!

PR #66, opened 2024-04-10
Refactor microagent MarketFunctions to take market_id as input

evangriffiths:
Changes to microagent to prevent the agent from having to pass the 'market question' around to various Functions

- Changed `GetMarkets` to return market questions and IDs
- Added `GetMarketProbability` for getting the market p_yes from the market id
- `BuyTokens` Functions take in market_id instead of market question
- Required a bump to PMAT 0.11.0
- Also added an integration test where the agent runs `GetMarketProbability`


prediction_market_agent/agents/known_outcome_agent/benchmark.py
```diff
@@ -39,7 +39,7 @@ def to_market(self) -> AgentMarket:
             ),
             volume=None,
             created_time=None,
-            close_time=None,
```
evangriffiths, line 42: Required change after bumping PMAT version

prediction_market_agent/agents/known_outcome_agent/deploy.py
```diff
@@ -33,7 +34,7 @@ class DeployableKnownOutcomeAgent(DeployableAgent):
     def load(self) -> None:
         self.markets_with_known_outcomes: dict[str, Result] = {}
 
-    def pick_markets(self, markets: list[AgentMarket]) -> list[AgentMarket]:
+    def pick_markets(self, markets: t.Sequence[AgentMarket]) -> list[AgentMarket]:
```
evangriffiths, line 37: Required change after bumping PMAT version

prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -63,11 +62,18 @@ def __init__(self, market_type: MarketType) -> None:
         self.market_type = market_type
         super().__init__()
 
+    @property
```
evangriffiths, line 65: Added helper function to `MarketFunction` class, as it's needed in a few places

prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -80,7 +86,27 @@ def __call__(self) -> list[str]:
...
+    def example_args(self) -> list[str]:
+        return [get_example_market_id(self.market_type)]
+
+    def __call__(self, market_id: str) -> list[str]:
+        return [
+            str(self.market_type.market_class.get_binary_market(id=market_id).p_yes)
+        ]
+
+
+class PredictPropabilityForQuestion(MarketFunction):
```
evangriffiths, line 109: Renamed otherwise it could easily be mistaken for `GetMarketProbability`

prediction_market_agent/agents/microchain_agent/utils.py
```diff
@@ -28,23 +29,23 @@
...
 
     def __str__(self) -> str:
-        return f"'{self.question}' with probability of yes: {self.p_yes:.2%}"
+        return f"'{self.question}', id: {self.id}"
 
 
 def get_binary_markets(market_type: MarketType) -> list[AgentMarket]:
     # Get the 5 markets that are closing soonest
     cls = market_type.market_class
-    markets: list[AgentMarket] = cls.get_binary_markets(
```
evangriffiths, line 47: Required change after bumping PMAT version

prediction_market_agent/agents/microchain_agent/functions.py
gabrielfior: I'm a bit confused - shouldn't this method be named something like `PlaceBet` or `BuyTokensForOutcome`? It's not really predicting anything, like the description says it allows the agent to purchase tokens.

tests/agents/microchain/test_functions.py
```diff
@@ -108,3 +109,10 @@ def test_engine_help(market_type: MarketType) -> None:
         engine.register(function(market_type=market_type))
 
     print(engine.help)
+
+
+@pytest.mark.parametrize("market_type", [MarketType.OMEN])
+def test_get_probability(market_type: MarketType) -> None:
+    market_id = "0x0020d13c89140b47e10db54cbd53852b90bc1391"
+    get_market_probability = GetMarketProbability(market_type=market_type)
+    get_market_probability(market_id)
```
gabrielfior, line 118: Is the assertion missing here?

tests/agents/microchain/test_agent.py
```diff
@@ -0,0 +1,87 @@
...
+
+    agent.bootstrap = ['Reasoning("I need to reason step-by-step")']
+    agent.run(iterations=5)
+
+    # history[-1] is 'user' stop message
+    # history[-2] is 'assistant' stop function call
+    m_json = json.loads(agent.history[-3]["content"])
+    m = MarketIDAndProbability.model_validate(m_json)
+    market: AgentMarket = market_type.market_class.get_binary_market(m.market_id)
+    assert market.p_yes == m.probability
```
gabrielfior, line 87: Like this test a lot!
Just wondering - `self.history` makes sense (I don't see another way to fetch microchain output - one could use `on_iteration_end` and assign a variable, but don't think it would be much better).

This is a lot hackier though when compared with crewai (`task.output.raw_content`, `task.output.json_output`, etc) and autogen. So probably something to keep in mind when comparing those frameworks.

prediction_market_agent/agents/known_outcome_agent/benchmark.py
```diff
@@ -39,7 +39,7 @@ def to_market(self) -> AgentMarket:
...
+            close_time=datetime.now() + timedelta(days=1),
```
kongzii, line 42:
```suggestion
            close_time=utcnow() + timedelta(days=1),
```

🙏

prediction_market_agent/agents/microchain_agent/functions.py
kongzii: @gabrielfior It's okay, just that Github hides the lines:

 
Screenshot by Dropbox Capture

tests/agents/microchain/test_agent.py
evangriffiths: Yeah that's a good point. I thought the same

tests/agents/microchain/test_functions.py
evangriffiths: Yes good point. I noticed that the way we calculate p_yes for closed omen markets is broken atm, so I made this task https://github.com/gnosis/prediction-market-agent-tooling/issues/181

I've added an assertion here that will fail once that task is resolved, but I think that's fine.

PR #67, opened 2024-04-11
Docker build on push to main

kongzii:


build_and_push.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# This script builds the docker image and pushes it to the GitHub Container Registry,
+# it should be done from Github's CD pipeline, but keeping it here for any testing purposes required in the future.
+docker build . -t ghcr.io/gnosis/prediction-market-agent:test && docker push ghcr.io/gnosis/prediction-market-agent:test
```
evangriffiths, line 4: Have you left this in just for manual testing?

kongzii: yeah so we can do it locally if needed, it has a special `test` tag at the end, so no conflicts with "prod" deployment

evangriffiths (approved):

PR #75, opened 2024-04-12
Implement PredictProbabilityForQuestion function for microchain agent

evangriffiths:
Two new Functions for microchain agent:

- `PredictProbabilityForQuestion` that calls the mech-client. However, from experimentation this is not very reliable. Currently it fails with timeout ~50% of the time. So while this is being debugged...
- `PredictProbabilityForQuestionLocal` that calls the mech function's python directly. This pulls in the mech repo as a git submodule.

## Summary by CodeRabbit

- **New Features**
	- Introduced new prediction functionalities in the market agent, including local and external tool interactions for enhanced probability predictions.
	- Added a new prediction agent with capabilities to generate research reports and utilize OpenAI models for market analysis.
- **Bug Fixes**
	- Increased the number of iterations in the prediction agent's run method for improved output accuracy.
- **Documentation**
	- Updated project dependencies and configurations in `pyproject.toml` and GitHub CI workflows.
- **Refactor**
	- Streamlined import statements and removed redundant code across various modules.
- **Chores**
	- Updated `.gitignore` to exclude specific files and added essential libraries to the project setup.


pyproject.toml
```diff
@@ -21,28 +21,30 @@ google-search-results = "*"
...
+prediction-market-agent-tooling = {git = "https://github.com/gnosis/prediction-market-agent-tooling.git", rev = "evan/mech-client-deps-compat"}
 pydantic-settings = "^2.1.0"
 autoflake = "^2.2.1"
 isort = "^5.13.2"
 markdownify = "^0.11.6"
 tavily-python = "^0.3.1"
 # TODO remove when PR is merged
 microchain-python = {git = "https://github.com/evangriffiths/microchain.git", rev = "evan/token-tracker"}
+types-requests = "2.31.0.0"
+crewai = {version = "0.11.0", extras = ["tools"]}
```
evangriffiths, line 45: crewai conflict with mech-client deps

pyproject.toml
```diff
@@ -21,28 +21,30 @@ google-search-results = "*"
...
-types-requests = "^2.31.0.20240106"
 black = "^23.12.1"
 google-cloud-functions = "^1.16.0"
 google-cloud-resource-manager = "^1.12.0"
 poetry = "^1.7.1"
 poetry-plugin-export = "^1.6.0"
 functions-framework = "^3.5.0"
 cron-validator = "^1.0.8"
-prediction-market-agent-tooling = "^0.13.0"
+prediction-market-agent-tooling = {git = "https://github.com/gnosis/prediction-market-agent-tooling.git", rev = "evan/mech-client-deps-compat"}
```
evangriffiths, line 36: TODO make mech-client a dependency of PMAT and deal with conflicts there

pyproject.toml
gabrielfior: I assume this will also be reverted back to v0.28 once PMAT adds mech-client as dependency?

prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -101,19 +105,31 @@ def __call__(self, market_id: str) -> list[str]:
...
-            return 0.41
-        if a == "Will Bitcoin hit 100k in 2024?":
-            return 0.22
+        return [get_example_market_id(self.market_type)]
 
-        return 0.0
+    def __call__(self, question: str) -> float:
+        private_key = APIKeys().bet_from_private_key.get_secret_value()
+        with saved_str_to_tmpfile(private_key) as tmpfile_path:
+            response = interact(
```
gabrielfior, line 122: This costs money to run. Should we do some kind of balance check here to see if the agent has enough funds? Otherwise this will throw an error.

pyproject.toml
kongzii: mech-client is super very strictly versioned. if they do this in other repos as well then it's even worse 🙈  I think 
it's worth a try to replace these `==` (exact version) with `^` (any newer version except for the versions with breaking changes). 

Maybe there still will be some conflicts but a lot less, at least hopefully.

https://github.com/valory-xyz/mech-client/blob/main/pyproject.toml#L27

Screenshot by Dropbox Capture

prediction_market_agent/agents/microchain_agent/functions.py
evangriffiths: Added. I don't think there's a way to query the cost of a mech call, so I've hard-coded it here. 

I tried playing around with the value [here](https://github.com/valory-xyz/mech-client/blob/main/mech_client/interact.py#L338) that sets the cost of the mech call, and it fails for anything <0.01xDai, and works for anything >=.

pyproject.toml
```diff
@@ -21,28 +21,29 @@ google-search-results = "*"
...
-prediction-market-agent-tooling = "^0.13.1"
+prediction-market-agent-tooling = {git = "https://github.com/gnosis/prediction-market-agent-tooling.git", rev = "evan/add-mech-client-dependency"}
 pydantic-settings = "^2.1.0"
 autoflake = "^2.2.1"
 isort = "^5.13.2"
 markdownify = "^0.11.6"
 tavily-python = "^0.3.1"
 # TODO remove when PR is merged
 microchain-python = {git = "https://github.com/evangriffiths/microchain.git", rev = "evan/token-tracker"}
+setuptools = "^69.5.1"
```
evangriffiths, line 46: Required by mech-client, although not a dependency. Will make a PR against the aea repo, but for now adding here

```
    from aea.crypto.plugin import load_all_plugins
  File "/Users/evan/dev/prediction-market-agent/.venv/lib/python3.10/site-packages/aea/crypto/plugin.py", line 26, in <module>
    from pkg_resources import EntryPoint, WorkingSet
ModuleNotFoundError: No module named 'pkg_resources'
```

prediction_market_agent/agents/known_outcome_agent/known_outcome_agent.py
```diff
@@ -141,26 +140,6 @@ def has_known_result(self) -> bool:
 """
 
 
-def completion_str_to_json(completion: str) -> dict[str, t.Any]:
```
evangriffiths, line 144: Moved to prediction_market_agent/utils.py

pyproject.toml
```diff
@@ -21,28 +21,30 @@ google-search-results = "*"
...
+jsonschema = "4.3.3"
```
evangriffiths, line 47: Same here

pyproject.toml
evangriffiths: Unfortunately doesn't help 😞 

The conflicts come from further down the dependency tree
- here: https://github.com/valory-xyz/open-aea/blob/1bb984ba44b028d941266ffeb13558bb4f32645d/plugins/aea-ledger-ethereum/setup.py#L46
- and here https://github.com/valory-xyz/open-aea/blob/1bb984ba44b028d941266ffeb13558bb4f32645d/setup.py#L63

So there's strictness to fix first in https://github.com/valory-xyz/open-aea and in the mech-client repo

.github/workflows/python_ci.yaml
```diff
@@ -25,17 +27,21 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          submodules: true
       - uses: ./.github/actions/python_prepare
       - name: Run pytest
-        run: poetry run pytest
+        run: poetry run pytest tests
```
evangriffiths, line 34: Don't pick up the tests from inside the mech submodule

prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -98,22 +105,80 @@ def __call__(self, market_id: str) -> list[str]:
...
+        tavily_api_key = keys.tavily_api_key.get_secret_value()
+        kwargs = {
+            "tool": "prediction-with-research-conservative",
+            "prompt": question,
+            "api_keys": {
+                "openai": openai_api_key,
+                "tavily": tavily_api_key,
+            },
+        }
+        response = prediction_with_research_report.run(**kwargs)
```
kongzii, line 179: Why **kwargs and not just 

```python
 prediction_with_research_report.run(
  tool="prediction-with-research-conservative",
  prompt=question,
  api_keys={
                "openai": openai_api_key,
                "tavily": tavily_api_key,
            },
)
```

?

kongzii (approved):
So because the mech calling over the internet doesn't work very well, this adds it as a submodule and use directly?

prediction_market_agent/agents/microchain_agent/functions.py
evangriffiths: Une bonne recommandation!

evangriffiths:
> So because the mech calling over the internet doesn't work very well, this adds it as a submodule and use directly?

Yes, that's right. I have yet to debug why it's unreliable.

PR #76, opened 2024-04-15
Install extras from PMAT

kongzii:
## Summary by CodeRabbit

- **New Features**
	- Updated the `prediction-market-agent-tooling` dependency to include additional extras like "langchain" and "google" for enhanced functionality.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ poetry = "^1.7.1"
 poetry-plugin-export = "^1.6.0"
 functions-framework = "^3.5.0"
 cron-validator = "^1.0.8"
-prediction-market-agent-tooling = "^0.13.1"
+prediction-market-agent-tooling = { version = "^0.13.1", extras = ["langchain", "google"] }
 pydantic-settings = "^2.1.0"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -35,7 +35,7 @@ poetry = "^1.7.1"
 poetry-plugin-export = "^1.6.0"
 functions-framework = "^3.5.0"
 cron-validator = "^1.0.8"
-prediction-market-agent-tooling = "^0.13.1"
+prediction-market-agent-tooling = { version = "^0.13.1", extras = ["langchain", "google"] }
```
evangriffiths, line 38: You can use `extras = ["all"]` to save specifying them manually
https://python-poetry.org/docs/pyproject/#extras

kongzii: it seems like it should work, but if I do it, it removes them from the lock file:

Screenshot by Dropbox Capture

and after removing and installing the poetry env again, it's not there:

Screenshot by Dropbox Capture

but if I keep it explicit it works:

Screenshot by Dropbox Capture

evangriffiths (approved):

PR #92, opened 2024-04-19
Allow specifying of multiple CLOSE_TIME_UP_TO_N_DAYS for replicator

kongzii:
## Summary by CodeRabbit

- **New Features**
	- Enhanced the market replication feature to dynamically adjust closure times based on different days, allowing for more tailored market operations.


prediction_market_agent/agents/replicate_to_omen_agent/deploy.py
```diff
@@ -26,7 +26,7 @@ class ReplicateSettings(BaseSettings):
 
     N_TO_REPLICATE: int
     INITIAL_FUNDS: str
-    CLOSE_TIME_UP_TO_N_DAYS: int
+    CLOSE_TIME_UP_TO_N_DAYS: list[int]
```
kongzii, line 29: So we can force to replicate markets that will close soon, but at the same time these that will close in a year or so

evangriffiths (approved):

PR #94, opened 2024-04-22
Move mech submodule, add test

evangriffiths:
## Summary by CodeRabbit

- **Refactor**
    - Streamlined handling of API keys and improved code maintainability in the prediction market agent.
- **New Features**
    - Introduced a new class for managing Google Search API keys securely.
    - Added utility functions for better interaction with the Mech prediction market tool.
- **Tests**
    - Implemented a new test to ensure the prediction market agent accurately predicts probabilities.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "prediction_market_agent/agents/microchain_agent/mech/mech"]
-	path = prediction_market_agent/agents/microchain_agent/mech/mech
+[submodule "prediction_market_agent/tools/mech/mech"]
+	path = prediction_market_agent/tools/mech/mech
 	url = https://github.com/valory-xyz/mech.git
--- a/prediction_market_agent/tools/mech/utils.py
+++ b/prediction_market_agent/tools/mech/utils.py
@@ -1,0 +1,92 @@
+import os
+import tempfile
+import typing as t
+from contextlib import contextmanager
+from enum import Enum
+
+from mech_client.interact import ConfirmationType, interact
+from pydantic import BaseModel
+
+from prediction_market_agent.tools.mech.api_keys import MechAPIKeys
+from prediction_market_agent.tools.mech.mech.packages.polywrap.customs.prediction_with_research_report import (
+    prediction_with_research_report,
+)
+from prediction_market_agent.tools.mech.mech.packages.valory.customs.prediction_request import (
+    prediction_request,
+)
+from prediction_market_agent.utils import completion_str_to_json
+
+
+class MechResult(BaseModel):
+    p_yes: float
+    p_no: float
+    confidence: float
+    info_utility: float
+
+
+@contextmanager
+def saved_str_to_tmpfile(s: str) -> t.Iterator[str]:
+    # Write the string to the temporary file
+    with tempfile.NamedTemporaryFile(delete=False) as tmp:
+        tmp.write(s.encode())
+
+    yield tmp.name
+
+    # Finally remove the temporary file
+    os.remove(tmp.name)
+
+
+class MechTool(str, Enum):
+    PREDICTION_WITH_RESEARCH_REPORT = "prediction-with-research-conservative"
+    PREDICTION_ONLINE = "prediction-online"
+
+
+def mech_request(question: str, mech_tool: MechTool) -> MechResult:
+    private_key = MechAPIKeys().bet_from_private_key.get_secret_value()
+    with saved_str_to_tmpfile(private_key) as tmpfile_path:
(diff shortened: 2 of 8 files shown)


.gitmodules
```diff
@@ -1,3 +1,3 @@
-[submodule "prediction_market_agent/agents/microchain_agent/mech/mech"]
-	path = prediction_market_agent/agents/microchain_agent/mech/mech
+[submodule "prediction_market_agent/tools/mech/mech"]
```
gabrielfior, line 1: Why is it advised to use submodules vs poetry dependencies?

prediction_market_agent/tools/mech/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    prediction_request,
+)
+from prediction_market_agent.utils import completion_str_to_json
+
+
+class MechResult(BaseModel):
+    p_yes: float
+    p_no: float
+    confidence: float
+    info_utility: float
```
gabrielfior, line 24: We have a similar class `OutcomePrediction` somewhere else in the code - please unify

prediction_market_agent/tools/mech/utils.py
```diff
@@ -0,0 +1,92 @@
...
+            prompt=question,
+            # Taken from https://github.com/valory-xyz/mech?tab=readme-ov-file#examples-of-deployed-mechs
+            agent_id=6,
+            private_key_path=tmpfile_path,
+            # To see a list of available tools, comment out the tool parameter
+            # and run the function. You will be prompted to select a tool.
+            tool=mech_tool.value,
+            confirmation_type=ConfirmationType.WAIT_FOR_BOTH,
+        )
+        del os.environ[mech_strategy_env_var]
```
gabrielfior, line 63: Why deleting this? I assume mech writes this, right?

prediction_market_agent/tools/mech/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    elif mech_tool == MechTool.PREDICTION_ONLINE:
+        response = prediction_request.run(
+            tool=mech_tool.value,
+            prompt=question,
+            api_keys={
+                "openai": keys.openai_api_key.get_secret_value(),
+                "google_api_key": keys.google_search_api_key.get_secret_value(),
+                "google_engine_id": keys.google_search_engine_id.get_secret_value(),
+            },
+        )
```
gabrielfior, line 87: My understanding was that, if the method is called `run_..._local`, that a local resource would be used and no costs would rise from it.
When you call `prediction_request` (from the mech package), what happens?

.gitmodules
evangriffiths: Ahh, when I first did this I thought that [mech](https://github.com/valory-xyz/mech) wasn't configured to be installed as a python package, but I am wrong!

It isn't released on pypi, but it is possible to install by adding `mech = { git = "https://github.com/valory-xyz/mech.git", branch = "main" }` to the pyproject.toml file.

evangriffiths: I tried switching to this, and it's not a clean install due to dependency issues. Spent an hour trying to fix these, but no luck!

With changes to [PMA](https://github.com/gnosis/prediction-market-agent/pull/95) and [PMAT](https://github.com/gnosis/prediction-market-agent-tooling/pull/213) dependencies I was blocked by an unresolvable conflict with crewai.

With changes the other way to the [mech](https://github.com/valory-xyz/mech/pull/218) repo I was blocked by conflicts with the mech-client repo (required by PMAT).

evangriffiths: Submodule is a workaround for this for now... but I'll make a ticket to fix the conflicts and replace the submodule with a package dependency.

evangriffiths: issue here: https://github.com/gnosis/prediction-market-agent/issues/97

prediction_market_agent/tools/mech/utils.py
evangriffiths: This env var is read inside the `interact` function. So we set it before calling. We unset it after so as not to pollute the env vars

prediction_market_agent/tools/mech/utils.py
evangriffiths: > no costs would rise from it.

Yes, no costs incurred via crypto wallet payments. But these functions generally have search and LLM api calls, so those costs are incurred.

If you call mechs via the client, you pay 0.01 in xdai and don't need openai/search api keys. If you run locally, you need to these api keys, but no crypto wallet.

gabrielfior (approved):

PR #47, opened 2024-04-03
Agent that thinks more thoroughly about question and considers possible outcomes

gabrielfior:
Closes #40 

## Summary by CodeRabbit

- **New Features**
	- Introduced benchmarking functionality for CrewAI agents in prediction markets.
	- Added a DeployableThinkThoroughlyAgent for selecting and betting on prediction markets.
	- New functionalities for creating outcomes, determining probabilities, and making decisions in prediction markets.

- **Refactor**
	- Improved organization by moving the `market_is_saturated` function to a utility module for better reusability.

PR #104, opened 2024-04-24
Upgrade PMAT and ThinkThoroughly agent

kongzii:
## Summary by CodeRabbit

- **New Features**
  - Enhanced security by introducing `PrivateCredentials` for handling private data across various agent functions.
  - Added new prompts and refined existing ones in the `think_thoroughly_agent` to improve market prediction interactions.
  - Enhanced the `agent_app.py` script to include cost tracking and display time and cost information during operations.

- **Bug Fixes**
  - Fixed formatting issues with dollar signs in logging messages.

- **Refactor**
  - Updated various agents to use renamed attributes and improved handling of market data.
  - Refactored methods related to scenario generation and market predictions in `think_thoroughly_agent`.

### DIFF
--- a/prediction_market_agent/agents/known_outcome_agent/benchmark.py
+++ b/prediction_market_agent/agents/known_outcome_agent/benchmark.py
@@ -32,7 +32,7 @@ class QuestionWithKnownOutcome(BaseModel):
             url=self.url if self.url else "",
             id=self.question,
             question=self.question,
-            p_yes=Probability(
+            current_p_yes=Probability(
                 self.result.to_p_yes()
                 if self.result != Result.KNOWN_UNKNOWABLE
                 else 0.5
@@ -53,7 +53,7 @@ class KnownOutcomeAgent(AbstractBenchmarkedAgent):
         model: str,
         max_tries: int,
     ) -> None:
-        self.model = model
+        self.model: str = model
         self.max_tries = max_tries
         super().__init__(agent_name=agent_name, max_workers=max_workers)
 
--- a/prediction_market_agent/agents/microchain_agent/functions.py
+++ b/prediction_market_agent/agents/microchain_agent/functions.py
@@ -1,6 +1,7 @@
 import typing as t
 
 from microchain import Function
+from prediction_market_agent_tooling.config import PrivateCredentials
 from prediction_market_agent_tooling.markets.agent_market import AgentMarket
 from prediction_market_agent_tooling.markets.data_models import Currency, TokenAmount
 from prediction_market_agent_tooling.markets.markets import MarketType
@@ -92,7 +93,11 @@ class GetMarketProbability(MarketFunction):
 
     def __call__(self, market_id: str) -> list[str]:
         return [
-            str(self.market_type.market_class.get_binary_market(id=market_id).p_yes)
+            str(
+                self.market_type.market_class.get_binary_market(
+                    id=market_id
+                ).current_p_yes
+            )
         ]
 
 
@@ -171,7 +176,9 @@ class BuyTokens(MarketFunction):
         self.outcome_bool = get_boolean_outcome(
             outcome=self.outcome, market_type=market_type
         )
-        self.user_address = MicrochainAPIKeys().bet_from_address
+        self.user_address = PrivateCredentials.from_api_keys(
+            MicrochainAPIKeys()
+        ).public_key
         super().__init__(market_type=market_type)
 
     @property
@@ -232,7 +239,9 @@ class SellTokens(MarketFunction):
             outcome=self.outcome,
             market_type=market_type,
         )
-        self.user_address = MicrochainAPIKeys().bet_from_address
+        self.user_address = PrivateCredentials.from_api_keys(
+            MicrochainAPIKeys()
+        ).public_key
         super().__init__(market_type=market_type)
 
     @property
@@ -314,7 +323,9 @@ class GetBalance(MarketFunction):
 
 class GetPositions(MarketFunction):
     def __init__(self, market_type: MarketType) -> None:
-        self.user_address = MicrochainAPIKeys().bet_from_address
+        self.user_address = PrivateCredentials.from_api_keys(
+            MicrochainAPIKeys()
+        ).public_key
         super().__init__(market_type=market_type)
 
     @property
@@ -329,7 +340,9 @@ class GetPositions(MarketFunction):
         return []
 
     def __call__(self) -> list[str]:
-        self.user_address = MicrochainAPIKeys().bet_from_address
+        self.user_address = PrivateCredentials.from_api_keys(
+            MicrochainAPIKeys()
+        ).public_key
         positions = self.market_type.market_class.get_positions(
             user_id=self.user_address
         )
--- a/prediction_market_agent/agents/microchain_agent/functions.py
+++ b/prediction_market_agent/agents/microchain_agent/functions.py
@@ -1,8 +1,8 @@
 import typing as t
 
 from microchain import Function
-from prediction_market_agent_tooling.config import PrivateCredentials
 from prediction_market_agent_tooling.benchmark.utils import OutcomePrediction
+from prediction_market_agent_tooling.config import PrivateCredentials
 from prediction_market_agent_tooling.markets.agent_market import AgentMarket
 from prediction_market_agent_tooling.markets.data_models import Currency, TokenAmount
 from prediction_market_agent_tooling.markets.markets import MarketType
@@ -178,10 +178,7 @@ class BuyTokens(MarketFunction):
         self.outcome_bool = get_boolean_outcome(
             outcome=self.outcome, market_type=market_type
         )
-        self.user_address = PrivateCredentials.from_api_keys(
-            MicrochainAPIKeys()
-        ).public_key
-        self.user_address = APIKeys().bet_from_address
+        self.user_address = PrivateCredentials.from_api_keys(APIKeys()).public_key
         super().__init__(market_type=market_type)
 
     @property
@@ -242,9 +239,7 @@ class SellTokens(MarketFunction):
             outcome=self.outcome,
             market_type=market_type,
         )
-        self.user_address = PrivateCredentials.from_api_keys(
-            MicrochainAPIKeys()
-        ).public_key
+        self.user_address = PrivateCredentials.from_api_keys(APIKeys()).public_key
         super().__init__(market_type=market_type)
 
     @property
@@ -326,9 +321,7 @@ class GetBalance(MarketFunction):
 
 class GetPositions(MarketFunction):
     def __init__(self, market_type: MarketType) -> None:
-        self.user_address = PrivateCredentials.from_api_keys(
-            MicrochainAPIKeys()
-        ).public_key
+        self.user_address = PrivateCredentials.from_api_keys(APIKeys()).public_key
         super().__init__(market_type=market_type)
 
     @property
@@ -343,9 +336,7 @@ class GetPositions(MarketFunction):
         return []
 
     def __call__(self) -> list[str]:
-        self.user_address = PrivateCredentials.from_api_keys(
-            MicrochainAPIKeys()
-        ).public_key
+        self.user_address = PrivateCredentials.from_api_keys(APIKeys()).public_key
         positions = self.market_type.market_class.get_positions(
             user_id=self.user_address
         )
--- a/prediction_market_agent/agents/think_thoroughly_agent/deploy.py
+++ b/prediction_market_agent/agents/think_thoroughly_agent/deploy.py
@@ -29,7 +29,7 @@ class DeployableThinkThoroughlyAgent(DeployableAgent):
                     break
             else:
                 logger.info(
-                    f"Market {market.url} is too saturated to bet on with p_yes {market.p_yes}."
+                    f"Market {market.url} is too saturated to bet on with p_yes {market.current_p_yes}."
                 )
 
         return picked_markets
@@ -40,9 +40,9 @@ class DeployableThinkThoroughlyAgent(DeployableAgent):
         result = CrewAIAgentSubquestions().answer_binary_market(market.question)
         return (
             True
-            if result.decision == "y"
+            if result and result.decision == "y"
             else False
-            if result.decision == "n"
+            if result and result.decision == "n"
             else should_not_happen()
         )
 
--- a/prediction_market_agent/agents/think_thoroughly_agent/deploy.py
+++ b/prediction_market_agent/agents/think_thoroughly_agent/deploy.py
@@ -14,6 +14,11 @@ from prediction_market_agent.agents.utils import market_is_saturated
 
 
 class DeployableThinkThoroughlyAgent(DeployableAgent):
+    model: str = "gpt-4-turbo-2024-04-09"
+
+    def load(self) -> None:
+        self.agent = CrewAIAgentSubquestions(model=self.model)
+
     def pick_markets(self, markets: t.Sequence[AgentMarket]) -> t.Sequence[AgentMarket]:
         # We simply pick 5 random markets to bet on
         picked_markets: list[AgentMarket] = []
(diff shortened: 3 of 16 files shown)


prediction_market_agent/agents/known_outcome_agent/benchmark.py
```diff
@@ -32,7 +32,7 @@ def to_market(self) -> AgentMarket:
             url=self.url if self.url else "",
             id=self.question,
             question=self.question,
-            p_yes=Probability(
+            current_p_yes=Probability(
```
kongzii, line 35: PR includes changes required because of newest PMAT version

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -18,135 +21,162 @@
...
 class CrewAIAgentSubquestions:
-    def __init__(self) -> None:
-        llm = self._build_llm()
-        self.researcher = Agent(
+    model: str = "gpt-4-turbo-2024-04-09"
+
+    def _get_current_date(self) -> str:
+        return utcnow().strftime("%Y-%m-%d")
+
+    def _get_researcher(self) -> Agent:
```
kongzii, line 43: I added multi-threading for predicting individual subquestions. They are independent of each other so it speeds it up a lot. But because of that, every prediction needs to create its own agent instances.

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
```diff
@@ -1,100 +1,104 @@
...
+You are asked about a SCENARIO, which can have either 'YES' or 'NO' outcomes.
+
+However, based on the SCENARIO, you should create a list of other possible SCENARIOS, these SCENARIOS should also have either 'YES' or 'NO' possible outcomes.
+
+Here are a few examples of SCENARIOs and then SCENARIOs derived from them:
+
+Example 1:
+- SCENARIO: Will Gnosis Pay reach 100,000 users by 2025?
+- Possible other SCENARIOs:
+  - Will Gnosis Pay have less than 25,000 users by 2025?
```
kongzii, line 23: Instead of generating outcomes (ie `Gnosis Pay reaches less than 25,000 users by 2025`), I changed it to generate questions like this.

Agents accept the questions as input, so to keep it consistent

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
```diff
@@ -1,100 +1,104 @@
...
-        {scenario}
-        """
-
-CREATE_OUTCOMES_FROM_SCENARIO_OUTPUT = """
-  A list containing multiple bullet points. Each bullet point should start with '-'.
-  Each bullet point should contain a possible outcome resulting from the
-  provided SCENARIO. The produced outcomes should be mutually exclusive, i.e. only one of them should be true whereas
-  the remaining ones should be false.
-  """
+CREATE_REQUIRED_CONDITIONS_PROMPT = """
```
kongzii, line 1: Another subquestion generation for things that needs to happen in order for the question to be True

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -157,43 +188,27 @@ def generate_final_decision(
         )
         return output
 
-    def answer_binary_market(self, question: str) -> ProbabilityOutput:
-        outcomes = self.split_research_into_outcomes(question)
+    def answer_binary_market(self, question: str) -> ProbabilityOutput | None:
+        hypothetical_scenarios = self.get_hypohetical_scenarios(question)
+        conditional_scenarios = self.get_required_conditions(question)
+
+        sub_predictions = par_generator(
```
gabrielfior, line 195: I saw your comment regarding parallel execution.
I think this goes in the right direction, but would advise to use the standard parallelism from crewai (https://docs.crewai.com/core-concepts/Tasks/#asynchronous-execution) instead of managing futures ourselves.

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
gabrielfior: Left a comment on parallelism

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
```diff
@@ -1,100 +1,104 @@
...
+
+Here are a few examples of SCENARIOs and then SCENARIOs derived from them:
+
+Example 1:
+- SCENARIO: Will Gnosis Pay reach 100,000 users by 2025?
+- Possible other SCENARIOs:
+  - Will Gnosis Pay have less than 25,000 users by 2025?
+  - Will Gnosis Pay reache more than 25,000 users but less than 50,000 users by 2025?
+  - Will Gnosis Pay reache more than 50,000 users but less than 75,000 users by 2025?
+  - Will Gnosis Pay reache 100,000 users or more by 2025?
```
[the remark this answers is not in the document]
kongzii: Oops 😄 Updated, thanks

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
gabrielfior: So you are basically generating additional scenarios than the ones generated before?
Can you give examples on which scenarios were generated before and now, to justify this change?

prediction_market_agent/agents/think_thoroughly_agent/deploy.py
```diff
@@ -14,6 +14,11 @@
 
 
 class DeployableThinkThoroughlyAgent(DeployableAgent):
+    model: str = "gpt-4-turbo-2024-04-09"
```
gabrielfior, line 17: Aren't costs too prohibitive? That's why I resorted back to gpt-3.5

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
gabrielfior: Keep in mind that their parallelism can be tricky - I had cases where just one task was resolved but it already called subsequential parts of the code.
So I would test it manually first and check which tasks can indeed be made parallel. Let me know if I can be of help.

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
kongzii: Kind of questions generated before can be seen in the linked issue, for example:

Screenshot by Dropbox Capture

And I screenshotted the kind of questions it generates now into Slack:

image

prediction_market_agent/agents/think_thoroughly_agent/deploy.py
kongzii: gpt-4-turbo is much better on costs than gpt-4 (that was enabled by default back then):

Screenshot by Dropbox Capture

Unfortunately, gpt-3.5-turbo doesn't work well here. So I'm afraid this is the cheapest choice that (seems to) work.

evangriffiths (commented):
I like the idea of adding `conditional_scenarios` to help the agent reason about the final p_yes. But I think it changes the reasoning required when making the final decision, and I think this should be at least documented, and maybe also expressed in the prompt / logic.

Before, the `hypothetical_scenarios` were supposed to cover the whole probability space for some event, e.g. outcomes `a1`, `a2` and `a3` for event `A`, where `a1` is the 'Yes' outcome to the original question, and `a2` and `a3` are the extra scenarios generated by the agent. Then
```
p(A=a1) + p(A=a2) + p(A=a3) = 1
```
I can't see this explicitly stated anywhere, but I thought that's what @gabrielfior was intending with the design! So then the reasoning the agent might do in `generate_final_decision`, when it adjusts its estimate of `p(A=a1)` (i.e. the final p_yes) would be to normalize them like `p(A=a1) / p(A=a1) + p(A=a2) + p(A=a3)`.

But now you're adding in `conditional_scenarios`, which are like outcomes of different random variables, e.g. `p(B=b1)`, `p(C=c1)`. And we're implicitly saying here that because `A` is conditional on `B` and `C`, `p(A=a1 | B=b1) < p(B=b1)` and `p(A=a1 | C=c1) < p(C=c1)`. This requires different reasoning what it does for the hypothetical scenarios.

All that said, if the experiments show that the results are good, and so the agent gets this, then I guess it's fine, and it might just be helpful to add this in the comments.

Another point: just taking into account `hypothetical_scenarios` by allowing `generate_prediction_for_one_outcome` to have a `None` return value, it could have a bad knock-on effect when it comes to `generate_final_decision`. As instead of erroring out because one scenario fails, the subset of scenarios that are left could no longer cover the whole of the probability space. So the p_yes normalization reasoning that the agent does in this step could be screwed up. However, this point isn't so clear when the hypothetical scenarios are combined with `conditional_scenarios`.

prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -173,7 +178,7 @@ def __init__(self, market_type: MarketType, outcome: str):
         self.outcome_bool = get_boolean_outcome(
             outcome=self.outcome, market_type=market_type
         )
-        self.user_address = APIKeys().bet_from_address
+        self.user_address = PrivateCredentials.from_api_keys(APIKeys()).public_key
```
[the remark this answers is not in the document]
kongzii: Yeah, I agree, it's cumbersome. But let's do that in another PR?

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -157,43 +188,27 @@ def generate_final_decision(
...
+            hypothetical_scenarios.scenarios + conditional_scenarios.scenarios,
+            lambda x: (x, self.generate_prediction_for_one_outcome(x)),
```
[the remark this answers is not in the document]
kongzii: That is for `ProcessPoolExecutor`, but the default executor is `DEFAULT_THREADPOOL_EXECUTOR = ThreadPoolExecutor(max_workers=5)` (no need of pickling in this case)

kongzii:
@evangriffiths 

> Before, the hypothetical_scenarios were supposed to cover the whole probability space for some event, e.g. outcomes a1, a2 and a3 for event A, where a1 is the 'Yes' outcome to the original question, and a2 and a3 are the extra scenarios generated by the agent. Then p(A=a1) + p(A=a2) + p(A=a3) = 1

This is an interesting idea to have, but I don't think it was the original intention in this agent, was it @gabrielfior?

Previously, one of the examples in the prompt was

```
Example 1:
        - SCENARIO: Will Gnosis Pay reach 100,000 users by 2025?
        - Answer: '''
            - Gnosis Pay reaches less than 25,000 users by 2025
            - Gnosis Pay reaches more than 25,000 users but less than 50,000 users by 2025
            - Gnosis Pay reaches more than 50,000 users but less than 75,000 users by 2025
            - Gnosis Pay reaches 100,000 users or more by 2025
 ```

And these don't add up to `1` either, or?

This agent is based on Martin's ideas that he had some time ago:

Screenshot by Dropbox Capture

> So then the reasoning the agent might do in generate_final_decision, when it adjusts its estimate of p(A=a1) (i.e. the final p_yes) would be to normalize them like p(A=a1) / p(A=a1) + p(A=a2) + p(A=a3).

Based on the prompt for the final decision (and this was modified just a little), giving to agent many different scenarios with their probabilities if fine, and the agent should make sense of it:

Screenshot by Dropbox Capture

> Another point: just taking into account hypothetical_scenarios by allowing generate_prediction_for_one_outcome to have a None return value, it could have a bad knock-on effect 

Agreed that it won't be optimal, but I thought that if the agent does ~8 predictions, and let's say that some of them fails, it's a shame to not give the final result. 

For example from the screenshot in slack 

image

If the question about hitting $600 fails, it shouldn't be so bad for the final decision.

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
gabrielfior: The only problem I see with current implementation is the high memory usage (having 1 agent per outcome, and if I'm not mistaken, 5-8 outcomes per question)?
With crew-ai each agent acts as Worker, hence it's less memory intensive. The downside is it takes longer, hence the tradeoff here RAM vs time of execution.

gabrielfior:
@kongzii regarding your [comment ](https://github.com/gnosis/prediction-market-agent/pull/104#issuecomment-2075419412) (I couldn't reply directly there)
-> The main point is the sum of probability of each outcome not summing to 1. That should be the case in theory, however since we are processing outcomes in parallel (and not passing the other probabilities to the agent), it's natural to expect it to deliver probabilities that together don't add up to 1.
-> The way I implemented back then (and is still the case now) is to have the final agent look at all tuples (outcome, probability) and make a final choice.

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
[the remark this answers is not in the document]
evangriffiths: ah okay

evangriffiths:
> This is an interesting idea to have, but I don't think it was the original intention in this agent, was it @gabrielfior?

I can't find this in writing on the [original PR](https://github.com/gnosis/prediction-market-agent/pull/47), except for this comment from me, so maybe I'm just continuing the dream I was having 😆  
https://github.com/gnosis/prediction-market-agent/pull/47#pullrequestreview-1980454248

> And these don't add up to 1 either, or?

I would say they do! What other option is there?

> I thought that if the agent does ~8 predictions, and let's say that some of them fails, it's a shame to not give the final result.

That's fair enough.

I do think there is something to the point about how an agent reasons about probabilities of 'different outcomes of one random variable' vs probabilities of 'different outcomes of conditional random variables', but I think overall this is a great improvement and I haven't got a good idea on what to do about it, so am happy to approve as-is

evangriffiths (approved):

prediction_market_agent/agents/microchain_agent/functions.py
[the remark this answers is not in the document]
kongzii: @coderabbitai Please create an issue in prediction-market-agent-tooling repository to remove `PrivateCredentials` and make its methods just part of APIKeys

kongzii:
> I would say they do! What other option is there?

I see now what you meant! There are options

- Gnosis Pay reaches less than 25,000 users by 2025
- Gnosis Pay reaches more than 25,000 users but less than 50,000 users by 2025
- Gnosis Pay reaches more than 50,000 users but less than 75,000 users by 2025
- Gnosis Pay reaches 100,000 users or more by 2025

However, the option `more than 75,000 but less than 100,000 users` is missing. However, I see the idea that if it were included, the users should sum up to 1.

The original GitHub issue was also about playing with different dates, etc., so that wouldn't sum up to 1.

But @coderabbitai, please create an issue to upgrade the Think Thoroughly agent, such that he will generate subquestions in batches, and inside of these batches, probabilities should sum up to 1. Also mention that it's needed to check performance of current version of this agent, and maybe deploy the new one as  an another agent, instead of updating the existing one. Thanks!

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
kongzii: Errors are fine. But should the other agent take any significant amount of memory? If they are working in parallel in both cases, they need to store the scraped web content, generations, etc. and otherwise, I hope there is not extra memory overload, let's try it out:

If I set `max_workers=1`, then:

Screenshot by Dropbox Capture

And `max_workers=5`, then:

Screenshot by Dropbox Capture

---

So I don't see any significant difference but memory profiling can be tricky, we will see if Kube goes OOM 🙈 

---

Btw the question was, `Will the sun rise tomorrow?`. Disappointing that there is 0.01% chance that it will not! 😄

PR #116, opened 2024-04-26
Fix and test docker building

kongzii:
## Summary by CodeRabbit

- **New Features**
  - Added a job to build and push Docker images in the CI pipeline.

- **Refactor**
  - Updated the Dockerfile to use a new base Python image for both the builder and runtime stages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Install Poetry and create venv in the builder step,
 # then copy the venv to the runtime image, so that the runtime image is as small as possible.
-FROM --platform=linux/amd64 python:3.10.14-slim-bookworm AS builder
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Install Poetry and create venv in the builder step,
 # then copy the venv to the runtime image, so that the runtime image is as small as possible.
-FROM --platform=linux/amd64 python:3.10.14-slim-bookworm AS builder
+FROM --platform=linux/amd64 python:3.10.14-bookworm AS builder
```
kongzii, line 3: slim version doesn't include `gcc` which is now required for poetry install

evangriffiths (approved):

PR #119, opened 2024-04-29
Use latest PMAT with new pick_markets strategy

kongzii:
## Summary by CodeRabbit

- **New Features**
  - Introduced a feature that checks for recent betting activity to optimize market selection.
  - Set a limit of participating in a maximum of 5 markets per run to enhance performance.

- **Enhancements**
  - Improved market picking logic to skip markets recently betted on, ensuring a diverse range of market engagements.

### DIFF
--- a/prediction_market_agent/agents/think_thoroughly_agent/deploy.py
+++ b/prediction_market_agent/agents/think_thoroughly_agent/deploy.py
@@ -1,41 +1,82 @@
-import random
 import typing as t
+from datetime import timedelta
 
 from loguru import logger
+from prediction_market_agent_tooling.config import PrivateCredentials
 from prediction_market_agent_tooling.deploy.agent import Answer, DeployableAgent
 from prediction_market_agent_tooling.markets.agent_market import AgentMarket
+from prediction_market_agent_tooling.markets.manifold.api import (
+    get_authenticated_user,
+    get_manifold_bets,
+    get_manifold_market,
+)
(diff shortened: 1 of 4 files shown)


kongzii:
@evangriffiths DeployableCoinFlipAgent and DeployableKnownOutcomeAgent have their custom `pick_markets` logic. Seemed like they shouldn't be updated to this, is that right?

prediction_market_agent/agents/think_thoroughly_agent/deploy.py
```diff
@@ -1,84 +1,19 @@
...
 from prediction_market_agent.agents.think_thoroughly_agent.think_thoroughly_agent import (
     CrewAIAgentSubquestions,
 )
-from prediction_market_agent.utils import APIKeys
 
 
 class DeployableThinkThoroughlyAgent(DeployableAgent):
     model: str = "gpt-4-turbo-2024-04-09"
-    max_markets_per_run = 5
+    bet_on_n_markets_per_run = 1
```
evangriffiths, line 12: to make it cheaper?

kongzii: yeah, so far, he hasn't proved himself to be so good to be so expensive 😄 or do you think we should just keep it?

evangriffiths: No I think that's okay - our openai bill is getting out of hand a bit, so I'll do the same for the mech agents

PR #120, opened 2024-04-30
Thumbnail generator

kongzii:


scripts/image_app.py
```diff
@@ -0,0 +1,45 @@
...
+)
+
+with st.spinner("Generating image..."):
+    response = OpenAI().images.generate(
+        model="dall-e-3",
+        prompt=prompt_template.format(market=market.question),
+        size="1024x1024",
+        quality="standard",
+        n=1,
+    )
```
gabrielfior, line 39: Since you just need the question as prompt and the market query is only needed for when there is no input, I would suggest
-> Create a DataFrame with the first 5 markets (sorted by CLOSING SOONEST, for ex) and offer this as default input, for people to get inspired - people can copy-paste one of the market titles into the input field

scripts/image_app.py
```diff
@@ -0,0 +1,45 @@
...
+
+market = (
+    [m for m in markets if m.question == question][0]
+    if not custom_question_input
+    # If custom question is provided, just take some random market and update its question.
+    else markets[0].model_copy(update={"question": question, "current_p_yes": 0.5})
+)
+
+with st.spinner("Generating image..."):
+    response = OpenAI().images.generate(
```
gabrielfior, line 33: Also, can we use `dall-e-2` since it's just a preview and thus save some $?

scripts/image_app.py
kongzii: I don't understand. Why they can't just select the market from the dropdown?

Screenshot by Dropbox Capture

scripts/image_app.py
kongzii: Nope, dall-e-2 doesn't really work here, for example:

image

vs. the same market in dall-e-2:

image

gabrielfior: That's impressive. Closing my own comment :laughing:

scripts/image_app.py
gabrielfior: That works for me, thanks for the screenshot.

PR #124, opened 2024-05-01
Remove mech submodule

evangriffiths:
... and replace the prediction tool that called the mech code with calls into the prediction-prophet library

Fixes https://github.com/gnosis/prediction-market-agent/issues/326


kongzii (approved):

prediction_market_agent/agents/microchain_agent/market_functions.py
```diff
@@ -105,23 +100,63 @@ def __init__(
...
+        )
+        prediction = prophet_make_prediction(
+            market_question=question,
+            additional_information=research,
+            engine=self.model,
+            temperature=0,
+            api_key=self.keys.openai_api_key,
+        )
+        if prediction.outcome_prediction is None:
+            return "Failed to make a prediction."
```
gabrielfior, line 141: I wonder if it's best to throw an error? Otherwise the agent might take this and interpret this somehow as `p_yes`.

prediction_market_agent/tools/mech/utils.py
```diff
@@ -62,12 +44,15 @@ class MechTool(str, Enum):
 
 
 def mech_request(question: str, mech_tool: MechTool) -> MechResponse:
-    private_key = MechAPIKeys().bet_from_private_key.get_secret_value()
+    private_key = APIKeys().bet_from_private_key.get_secret_value()
```
gabrielfior, line 47: Is this function still used somewhere? My understanding is that the `general_agent` is not relying on the Mech for doing predictions, correct?

gabrielfior:
@evangriffiths I expected this PR to completely delete the `mech` folder from the repo (thus avoiding repetitive `mypy` and other CI issues) - is that the case?

evangriffiths:
> @evangriffiths I expected this PR to completely delete the `mech` folder from the repo (thus avoiding repetitive `mypy` and other CI issues) - is that the case?

Yep it's gone

prediction_market_agent/tools/mech/utils.py
evangriffiths: Yes, it's still used by `PredictProbabilityForQuestionMech` (see [here](https://github.com/gnosis/prediction-market-agent/blob/evan/remove-mech/prediction_market_agent/agents/microchain_agent/market_functions.py#L146-L150)), which itself is not used by the general agent atm, but I didn't want to delete the code until https://github.com/gnosis/prediction-market-agent/issues/327 was done/closed as wont-fix

prediction_market_agent/agents/microchain_agent/market_functions.py
evangriffiths: Yes good point. It looks like there's no reason for prophet to return a None prediction. Microchain has retry logic for errors too, so if we raise then the agent can retry the prediction.

PR #125, opened 2024-05-02
Fix sqlite3 on streamlit cloud

kongzii:


prediction_market_agent/tools/custom_crewai_tools.py
```diff
@@ -1,3 +1,7 @@
+from prediction_market_agent.utils import patch_sqlite3
+
+patch_sqlite3()
```
evangriffiths, line 3: Should this not be in the streamlit app instead of in the tools?

kongzii: Yes, that makes more sense! Moved.

evangriffiths (approved):

PR #126, opened 2024-05-02
Update ThinkThoroughly agent

kongzii:


prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -179,27 +195,38 @@ def generate_final_decision(
...
             logger.info(
-                f"'{scenario}' has prediction {prediction.p_yes * 100:.2f}% chance of being True, because: '{prediction.reasoning}'"
+                f"Starting to generate predictions for each scenario, iteration {iteration + 1} / {n_iterations}."
             )
 
+            sub_predictions = par_generator(
+                hypothetical_scenarios.scenarios + conditional_scenarios.scenarios,
+                lambda x: (
+                    x,
+                    self.generate_prediction_for_one_outcome(x, scenarios_with_probs),
```
kongzii, line 214: Curious what do you think about this. The idea is to feed into the prompt previous-iteration predictions to "normalise" them against each other.

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
```diff
@@ -67,6 +67,14 @@
 {sentence}
 Search and scrape the web for information that will help you give a high quality, nuanced answer to the question.
 """
+RESEARCH_OUTCOME_WITH_PREVIOUS_OUTPUTS_PROMPT = """
+Research and report on the following sentence:
+{sentence}
+Search and scrape the web for information that will help you give a high quality, nuanced answer to the question.
+
+You can use the following estimates, but you should not take them as the ground truth, they were generated independently and can be incorrect:
+{previous_scenarios_with_probabilities}
```
gabrielfior, line 76: I couldn't find the function which retrieves `previous_scenarios_with_probabilities` - I imagined this as some kind of memory, is this correct?

kongzii: No specific function, just previous predictions: https://github.com/gnosis/prediction-market-agent/pull/126/files#diff-02d2a6be16ea664e6d4128a79a4d9886a61b3f88c791d3b0bf786a809e720426R214

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
gabrielfior: I need more context into the changes to make up my mind.
Do I understand correctly that now you do 2 runs
-> 1st run: do as before, generate sub-predictions with probabilities that don't add to 1
-> 2nd run: pass in all sub-predictions to each agent, so that it can take those non-adding-to-1 probabilities in consideration

Do I understand the above correctly?
If so, I think a better approach would be to have an additional agent (e.g. `NormalizerAgent`), that takes a list of `sub_predictions` and simply normalizes them. This is because, in the current scenario, each agent is tasked with "normalizing" the probability by itself, so I think having only one agent accomplishing this is more consistent.

Curious about what you think.

kongzii: > that takes a list of sub_predictions and simply normalizes them

I was thinking about this too, but let's say the agent receives the following predictions:

- Will GNO hit $1000? p_yes=70%
- Will GNO hit $2000? p_yes=80%

What should the normalisation agent do? Increase the first one to >80% or decrease the second one to <70%? 

Both of the options are just guesses without additional information. So that's why I thought that just providing a list of previous predictions and making the research again should be better.

However, something like a normalisation agent that also does the research could be part of https://github.com/gnosis/prediction-market-agent/issues/108.

gabrielfior: Yes, happy with adding the normalization on #108 .
Approving this.

PR #129, opened 2024-05-02
Adds Long-Term-memory to microchain agent

gabrielfior:
- Add new function (`SummarizeLearnings`) for microchain agent to be able to retrieve previous actions
- Added `DB_storage` class for handling database interactions (with SQLModel under-the-hood)
- Wrote tests


prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -24,6 +26,12 @@ def main(
     for function in OMEN_FUNCTIONS:
         engine.register(function())
 
+    # This description below serves to unique identify agent entries on the LTM, and should be
+    # unique across instances (i.e. markets).
+    unique_task_description = f"microchain-agent-demo"
```
kongzii, line 31: If it's supposed to be unique across markets, shouldn't there be some market identifier in it?

prediction_market_agent/utils.py
```diff
@@ -8,6 +8,18 @@
...
+
+class DBKeys(BaseSettings):
+    model_config = SettingsConfigDict(
+        env_file=".env", env_file_encoding="utf-8", extra="ignore"
+    )
+    SQLALCHEMY_DB_URL: t.Optional[str] = None
+
+    @property
+    def sqlalchemy_db_url(self) -> str | None:
+        return self.SQLALCHEMY_DB_URL
```
kongzii, line 22: Is this property required if it only renames `SQLALCHEMY_DB_URL` to `sqlalchemy_db_url`?

prediction_market_agent/db/db_storage.py
```diff
@@ -0,0 +1,61 @@
...
+            sqlalchemy_db_url = keys.sqlalchemy_db_url
+
+        self.engine = create_engine(sqlalchemy_db_url)
+
+    def _initialize_db(self) -> None:
+        """
+        Creates the tables if they don't exist
+        """
+
+        # trick for making models import mandatory - models must be imported for metadata.create_all to work
```
evangriffiths, line 29: Is this necessary? You are doing `from prediction_market_agent.db.models import LongTermMemories` above.

prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -343,10 +330,30 @@ def __call__(self) -> list[str]:
...
+        """
+
+    @property
+    def example_args(self) -> list[str]:
+        return []
+
+    def __call__(self) -> t.Sequence[LongTermMemories]:
+        return self.long_term_memory.search()
+
+
```
evangriffiths, line 353: Not sure on the naming - it looks like `SummarizeLearnings` here is "remember old memories" as apposed to "making new memories", when I thought it would be the other way around. Either way, I think we need both.

I imagined this to look more like:
- Rename `SummarizeLearning` -> `SaveLearningSummary`. Instead of returning the learning as a string, it calls `long_term_memory.save_history(summary)`. And then in microchain_agent.py you don't have to call this manually - the management of saving and loading memory is offloaded to the agent.
- Add a new function `Load/RememberPastLearnings` that does what you're doing in this new `SummarizeLearnings` Function (i.e. `self.long_term_memory.search()`)

evangriffiths (commented):
Nice! Quite a clean and straight forward implementation. Just a few comments. Also, do you have any observations from using it? Can the agent make any use of it?

Also curious to hear at what point you think it might be worth moving from microchain to autogen/crewai since those frameworks already support different forms of memory.

prediction_market_agent/db/db_storage.py
gabrielfior: If I remove this, isort or mypy removes the import and it creates no tables.
Other option is to do `import *`, which mypy does not like.
Hence cheating here and logging this.

prediction_market_agent/agents/microchain_agent/functions.py
gabrielfior: -> SummarizeLearning -> SaveLearnings: I would rather keep the current implementation, since `agent.history` is more of a `microchain` feature than an agent feature. Also, other frameworks like Langchain manage memory outside of the agent, not as a function, hence would suggest we keep the current logic.
-> Accept renaming.

prediction_market_agent/utils.py
gabrielfior: Removed

prediction_market_agent/agents/microchain_agent/microchain_agent.py
gabrielfior: Yes, market_identifier was missing, thanks

gabrielfior:
> Nice! Quite a clean and straight forward implementation. Just a few comments. Also, do you have any observations from using it? Can the agent make any use of it?

I ran 2 iterations and the agent and it never called it. We should probably refactor the prompt for it to use it more often.

> Also curious to hear at what point you think it might be worth moving from microchain to autogen/crewai since those frameworks already support different forms of memory.

It's definitely worth it to move to autogen/crewai, we are already hitting the wall with the current functionality. I'm curious about autogen and think we should explore that, since we already have the `ThinkThoroughly` agent based on crew-ai. Let's chat more on Monday.

prediction_market_agent/agents/microchain_agent/functions.py
evangriffiths: > Also, other frameworks like Langchain manage memory outside of the agent, not as a function
Can you link to an example? I guess I can see why you would want writing to memory to be automated to make it reliable. But I would have thought reading from memory should be controlled by the agent. From how I see it the goal of giving the agent long-term memory specifically, is that it can learn to optimize the contents of its context so that it improves over time, and I think that needs to be given to the agent. Happy to be shown wrong though!

gabrielfior: As discussed today, idea is to write reasoning to DB in an automated way (at the end of the script) for more reliable storage.

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -2,19 +2,19 @@
...
-from prediction_market_agent.agents.microchain_agent.microchain_agent import get_agent
+from prediction_market_agent.agents.microchain_agent.microchain_agent import build_agent
 
 
 class DeployableMicrochainAgent(DeployableAgent):
     model = "gpt-4-1106-preview"
     n_iterations = 50
 
-    def run(self, market_type: MarketType) -> None:
+    def run(self, market_type: MarketType, _place_bet: bool = False) -> None:
```
evangriffiths, line 12: How come you've added this `_place_bet` argument? Looks like it's unused

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -1,4 +1,5 @@
 import typer
+from functions import MARKET_FUNCTIONS, MISC_FUNCTIONS, RememberPastLearnings
```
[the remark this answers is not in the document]
evangriffiths: Almost. Instead move `RememberPastLearnings` to below import (`from prediction_market_agent_tooling.markets.markets ...`), and remove others

prediction_market_agent/agents/microchain_agent/memory.py
```diff
@@ -0,0 +1,22 @@
+# inspired by crewAI's LongTermMemory (https://github.com/joaomdmoura/crewAI/blob/main/src/crewai/memory/long_term/long_term_memory.py)
+from typing import Any, Dict, Sequence
+
+from prediction_market_agent.db.db_storage import DBStorage
+from prediction_market_agent.db.models import LongTermMemories
+
+
+# In the future, create a base class which this class extends.
+class LongTermMemory:
+    def __init__(self, task_description: str, storage: DBStorage):
```
evangriffiths, line 10: The only time this is called in this PR, it's called with `storage=DBStorage()`. If that's the only way we expect it to be called, we could just remove the arg, and initialize that here with `self.storage = DBStorage()`

prediction_market_agent/agents/microchain_agent/deploy.py
gabrielfior: It's unused but mypy complains that the subclass is not implementing parent class correctly - is it not the case with you?

kongzii: Do you have the latest main changes? `_place_bet` was removed in the latest PMAT.

prediction_market_agent/agents/microchain_agent/memory.py
gabrielfior: It is the only way I can imagine at this time, so implementing your suggestion

evangriffiths (approved):

PR #132, opened 2024-05-03
Add microchain agent app

evangriffiths:
Allows the user to interact with the agent via iterative execution of the agent for a configurable number of steps, interleaved with user-input reasoning.

Screenshot 2024-05-03 at 09 54 18

### DIFF
--- a/prediction_market_agent/agents/microchain_agent/README.md
+++ b/prediction_market_agent/agents/microchain_agent/README.md
@@ -1,0 +1,19 @@
+# Microchain Agent
+
+This agent participates in prediction markets with a high degree of autonomy. It is equipped with tools to access the prediction market APIs, and can use its own reasoning to guide its betting strategy.
+
+## Useage
+
+Run the agent in the terminal via:
+
+```bash
+python prediction_market_agent/agents/microchain_agent/microchain_agent.py
+```
+
+Use `--help` to see optional arguments.
+
+Or run interactively in a Streamlit browser app via:
+
+```bash
+streamlit run prediction_market_agent/agents/microchain_agent/app.py
+```
--- a/prediction_market_agent/agents/microchain_agent/README.md
+++ b/prediction_market_agent/agents/microchain_agent/README.md
@@ -2,7 +2,7 @@
 
 This agent participates in prediction markets with a high degree of autonomy. It is equipped with tools to access the prediction market APIs, and can use its own reasoning to guide its betting strategy.
 
-## Useage
+## Usage
 
 Run the agent in the terminal via:
 
--- a/prediction_market_agent/agents/microchain_agent/app.py
+++ b/prediction_market_agent/agents/microchain_agent/app.py
@@ -17,7 +17,7 @@ from prediction_market_agent_tooling.tools.costs import openai_costs
 from prediction_market_agent.agents.microchain_agent.microchain_agent import get_agent
 
 
-def check_api_keys():
+def check_api_keys() -> None:
     keys = APIKeys()
     if not keys.OPENAI_API_KEY:
         st.error("No OpenAI API Key provided via env var/secret.")
@@ -27,16 +27,15 @@ def check_api_keys():
         st.stop()
 
 
-def run_agent(agent: Agent, iterations: int):
+def run_agent(agent: Agent, iterations: int) -> None:
     with openai_costs(model) as costs:
         with st.spinner("Agent is running..."):
             for _ in range(iterations):
                 agent.run(iterations=1, resume=True)
-                # st.session_state.is_first_run = False
         st.session_state.running_cost += costs.cost
 
 
-def execute_reasoning(agent: Agent, reasoning: str):
+def execute_reasoning(agent: Agent, reasoning: str) -> None:
     with openai_costs(model) as costs:
         agent.execute_command(f'Reasoning("{reasoning}")')
         st.session_state.running_cost += costs.cost
@@ -55,7 +54,6 @@ if "agent" not in st.session_state:
     st.session_state.agent.bootstrap = [f'Reasoning("I need to reason step by step")']
     st.session_state.agent.reset()
     st.session_state.agent.build_initial_messages()
-    # st.session_state.is_first_run = True
     st.session_state.running_cost = 0.0
 
 user_reasoning = st.text_input("Reasoning")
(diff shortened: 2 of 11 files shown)


prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def check_api_keys() -> None:
+    keys = APIKeys()
+    if not keys.OPENAI_API_KEY:
+        st.error("No OpenAI API Key provided via env var/secret.")
+        st.stop()
+    elif not keys.BET_FROM_PRIVATE_KEY:
+        st.error("No wallet private key provided via env var/secret.")
+        st.stop()
```
kongzii, line 27: This is handy. We could even have something like `def check_all_required_keys(list_of_keys_that_are_required_for_given_app)`. But that's fine for another PR, I'm just commenting on an idea here.

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def run_agent(agent: Agent, iterations: int) -> None:
+    with openai_costs(model) as costs:
```
kongzii, line 31: Should this be taken from global variable?

prediction_market_agent/utils.py
```diff
@@ -96,3 +97,25 @@ def patch_sqlite3() -> None:
...
+            return asyncio.get_event_loop()
+        except RuntimeError as ex:
+            if "There is no current event loop in thread" in str(ex):
+                loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(loop)
+                return asyncio.get_event_loop()
+            else:
+                raise ex
+
+    loop = get_or_create_eventloop()
```
kongzii, line 119: Seems like this line of code is never achieved.

try will return, `if` in the except block will return and `else` in the except block will raise.

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -0,0 +1,22 @@
...
+    def run(self, market_type: MarketType, _place_bet: bool = True) -> None:
+        """
+        Override main 'run' method, as the all logic from the helper methods
+        is handed over to the agent.
+        """
+        agent: Agent = get_agent(
+            market_type=market_type,
+            model=self.model,
+        )
+        agent.bootstrap = [f'Reasoning("I need to reason step by step")']
```
kongzii, line 21: Why is the `.bootstrap = ` always separate from `get_agent` or from the agent's initialisation? Feels like it could be just one step. (but even if yes, no need to edit in this PR)

prediction_market_agent/agents/microchain_agent/README.md
```diff
@@ -0,0 +1,19 @@
+# Microchain Agent
```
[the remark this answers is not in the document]
evangriffiths: That's literally the same?

prediction_market_agent/agents/microchain_agent/app.py
evangriffiths: Good idea. I've made a new file, streamlit_utils.py and put `check_required_api_keys` in there (and moved `streamlit_asyncio_event_loop_hack` alongside it)

prediction_market_agent/agents/microchain_agent/app.py
evangriffiths: Ah good spot. I've added a model arg to this function

prediction_market_agent/utils.py
evangriffiths: Not sure I understand. Here it's calling the inline function above, so this is the first line that runs

prediction_market_agent/agents/microchain_agent/deploy.py
evangriffiths: Good point. Have moved `.bootstrap =` inside `get_agent`

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -0,0 +1,108 @@
...
+    elif not keys.BET_FROM_PRIVATE_KEY:
+        st.error("No wallet private key provided via env var/secret.")
+        st.stop()
+
+
+def run_agent(agent: Agent, iterations: int, model: str) -> None:
+    with openai_costs(model) as costs:
+        with st.spinner("Agent is running..."):
+            for _ in range(iterations):
+                agent.run(iterations=1, resume=True)
```
gabrielfior, line 36: When calling agent.run(iterations=1), I'm worried the history (agent.history) is overwritten, i.e. it has no idea of the previous actions it took earlier.
So you might want to track the agent.history via an external variable, e.g. 
```
temp.history = []
for _ in range(iterations):
    agent.run(iterations=1) # ToDo - Add history to agent (note that it will be added to the prompt)
    temp_history = agent.history
```

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -9,39 +9,60 @@
...
-    
-    """
-
-    agent.bootstrap = ['Reasoning("I need to reason step-by-step")']
-    agent.run(iterations=10)
+    agent.prompt = SYSTEM_PROMPT.format(engine_help=engine.help)
+    agent.bootstrap = [
+        'Reasoning("I need to reason step by step. Start by assessing my '
+        'current position and balance.")'
+    ]
```
gabrielfior, line 47: Should we also add some explanation about how to make money in such a market?
Maybe that's the reason why it just bets everything in a simple market, it has no idea what it's doing

pyproject.toml
```diff
@@ -41,7 +41,7 @@ autoflake = "^2.2.1"
 isort = "^5.13.2"
 markdownify = "^0.11.6"
 tavily-python = "^0.3.1"
-microchain-python = "^0.3.5"
+microchain-python = { git = "https://github.com/galatolofederico/microchain.git", branch = "main" }
```
gabrielfior, line 44: Sorry, why has this been changed?

prediction_market_agent/agents/microchain_agent/app.py
evangriffiths: The `resume=True` flag means that history isn't overwritten (see this microchain PR here https://github.com/galatolofederico/microchain/pull/10)

prediction_market_agent/agents/microchain_agent/microchain_agent.py
evangriffiths: Yeah maybe. Need to have more of a play around with it to see what performs best. I don't think we need to be too worried about this for now. Btw I added a failsafe `MAX_AMOUNT` to `BuyTokens` so it can't do that now :D

prediction_market_agent/agents/microchain_agent/app.py
gabrielfior: Excellent, thanks

prediction_market_agent/agents/microchain_agent/microchain_agent.py
gabrielfior: works for me

gabrielfior (approved):

prediction_market_agent/utils.py
kongzii: Ah, I didn't notice that that try-except code is inside of another function! 😄 🤦

prediction_market_agent/agents/microchain_agent/README.md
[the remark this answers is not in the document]
kongzii: Code rabbit, there is no need to create artificial work for yourself; we know you can be useful.

PR #147, opened 2024-05-07
Microchain streamlit app improvements

evangriffiths:
Two small improvements:

1. Display the agent's history after each iteration, as apposed to at the end of all iterations
2. Grey out the `model` selection box once the agent has been initialized (once the first run is triggered)

### DIFF
--- a/prediction_market_agent/agents/microchain_agent/app.py
+++ b/prediction_market_agent/agents/microchain_agent/app.py
@@ -13,27 +13,16 @@ from prediction_market_agent_tooling.tools.utils import check_not_none
 
 from prediction_market_agent.agents.microchain_agent.microchain_agent import get_agent
 from prediction_market_agent.agents.microchain_agent.utils import (
+    get_initial_history_length,
     has_been_run_past_initialization,
 )
 from prediction_market_agent.streamlit_utils import check_required_api_keys
-from prediction_market_agent.utils import APIKeys
-
-
-def check_api_keys() -> None:
-    keys = APIKeys()
-    if not keys.OPENAI_API_KEY:
-        st.error("No OpenAI API Key provided via env var/secret.")
-        st.stop()
-    elif not keys.BET_FROM_PRIVATE_KEY:
-        st.error("No wallet private key provided via env var/secret.")
-        st.stop()
 
 
 def run_agent(agent: Agent, iterations: int, model: str) -> None:
     with openai_costs(model) as costs:
         with st.spinner("Agent is running..."):
             for _ in range(iterations):
-                # TODO with spinner...
                 agent.run(iterations=1, resume=True)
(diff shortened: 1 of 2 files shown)


prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -40,9 +30,48 @@ def run_agent(agent: Agent, iterations: int, model: str) -> None:
...
+
+def agent_is_initialized() -> bool:
+    return "agent" in st.session_state
+
+
+def maybe_initialize_agent(model: str) -> None:
+    # Initialize the agent
+    if not agent_is_initialized():
+        st.session_state.agent = get_agent(market_type=MarketType.OMEN, model=model)
+        st.session_state.agent.reset()
```
gabrielfior, line 67: This might be the wrong place, but I see a discrepancy between the model used and the model logged.

1. Model selected on streamlit

image


2 - Model logged

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -40,9 +30,48 @@ def run_agent(agent: Agent, iterations: int, model: str) -> None:
...
+def display_all_history(agent: Agent) -> None:
+    """
+    Display the agent's history in the Streamlit app.
+    """
+    # Skip the initial messages
+    history = agent.history[get_initial_history_length(agent) :]
+
+    for h in history:
+        st.chat_message(h["role"]).write(h["content"])
+
```
gabrielfior, line 46: I see some errors in the console (e.g. function called with wrong number of args) that don't pop-up in the dashboard.
Maybe we can display those if the user toggles a "Debug" button or similar?

image

prediction_market_agent/agents/microchain_agent/app.py
evangriffiths: Ah yes, this `ENGINE: gpt-4...` logging comes from the local mech code. You can't universally pass in a model arg there, so I left it to default behaviour. Once https://github.com/gnosis/prediction-market-agent/pull/124 is merged this issue will go away

prediction_market_agent/agents/microchain_agent/app.py
evangriffiths: Yeah that's annoying - I'm not sure how that would work cleanly with microchain. Those statements are `print`ed (e.g. here https://github.com/galatolofederico/microchain/blob/main/microchain/engine/agent.py#L27), not logged or stored in the agent's history.

gabrielfior: Ow I see - it would be nice to see the "complete" history including errors. Let's try to achieve that on the autogen one!

gabrielfior (approved):

PR #179, opened 2024-05-14
Evan/microchain gpt3.5 streamlit app deployment

evangriffiths:


prediction_market_agent/db/models.py
```diff
@@ -6,6 +6,7 @@
 
 class LongTermMemories(SQLModel, table=True):
     __tablename__ = "long_term_memories"
+    __table_args__ = {"extend_existing": True}
```
evangriffiths, line 9: Fixes `sqlalchemy.exc.InvalidRequestError: Table 'long_term_memories' is already defined for this MetaData instance.`. Is it a proper fix?

kongzii:
After this is merged, it should work to uncomment this block to deploy this in Kube: https://github.com/gnosis/gnosisai-infrastructure/pull/47/files#diff-da6cf1ab9cea6ebc8230d286607325e9adcad66ab763b842a1d974ec809c4b8eR80

PR #180, opened 2024-05-14
Misc microchain agent improvements and fixes

evangriffiths:


prediction_market_agent/db/models.py
```diff
@@ -6,6 +6,7 @@
 
 class LongTermMemories(SQLModel, table=True):
     __tablename__ = "long_term_memories"
+    __table_args__ = {"extend_existing": True}
```
evangriffiths, line 9: Fixes `sqlalchemy.exc.InvalidRequestError: Table 'long_term_memories' is already defined for this MetaData instance.`

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -56,10 +57,13 @@ def build_agent(
         temperature=0.7,
     )
     agent = Agent(llm=LLM(generator=generator), engine=engine)
+    agent.max_tries = 3
```
evangriffiths, line 60: Default `max_tries` is 10, which is excessive (leaves the user waiting a long time before feedback)

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -25,6 +25,7 @@
 {engine_help}
 
 Only output valid Python function calls.
+Make 'Reasoning' calls frequently - at least every other call.
```
evangriffiths, line 28: Instead of just making `Buy...`, `Sell...`,`Predict...` calls in succession, it intersperses these with reasoning. Now more 'chain-of-thought'-y

prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -191,8 +191,9 @@ def description(self) -> str:
         return (
             f"Use this function to buy {self.outcome} outcome tokens of a "
             f"prediction market. The first parameter is the market id. The "
-            f"second parameter specifies how much {self.currency} you spend."
-            f"This is capped at {self.MAX_AMOUNT}{self.currency}."
+            f"second parameter specifies the VALUE of tokens to sell in "
+            f"{self.currency}. This is NOT the same as the number of tokens."
+            f"If this fails, try selling a smaller amount."
```
evangriffiths, line 196: Agent would always start buying new tokens. Now sells too.

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -96,7 +101,7 @@ def get_market_function_bullet_point_list() -> str:
     if not agent_is_initialized():
         model = st.selectbox(
             "Model",
-            ["gpt-4-turbo-2024-04-09", "gpt-3.5-turbo-0125"],
+            ["gpt-3.5-turbo-0125", "gpt-4o-2024-05-13"],
```
evangriffiths, line 104: default to gpt3.5. We're mostly still just experimenting, so it's preferable to run a cheaper model for most runs. Saves on fewer clicks. Also, switch gpt4 option to new `4o`.

prediction_market_agent/agents/microchain_agent/functions.py
gabrielfior: Does `sell` here mean `selling xDAI for outcome tokens`? If yes, this is correct, but I found it a bit confusing.
What do you think of rephrasing along the lines of "If fails, try purchasing a smaller amount of outcome tokens" or similar?
I think phrasing it in terms of `outcome tokens` is easier for the agent to understand - what do you think?

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -56,10 +57,13 @@ def build_agent(
...
     )
     agent = Agent(llm=LLM(generator=generator), engine=engine)
+    agent.max_tries = 3
     agent.prompt = SYSTEM_PROMPT.format(engine_help=engine.help)
     agent.bootstrap = [
         'Reasoning("I need to reason step by step. Start by assessing my '
-        'current position and balance.")'
+        "current positions and balance. Do I have any positions in the markets "
+        "returned from GetMarkets? Consider selling overvalued tokens AND "
+        'buying undervalued tokens.")'
```
gabrielfior, line 66: That's great - nudging in the right direction seems like a great approach

Another thing we could explore is to provide more concrete examples with actual numbers, like on AutoTx (https://github.com/polywrap/AutoTx/blob/main/autotx/agents/SendTokensAgent.py#L23)

kongzii (approved):

prediction_market_agent/agents/microchain_agent/microchain_agent.py
evangriffiths: Good idea. Will leave to later a PR if that's okay

prediction_market_agent/agents/microchain_agent/functions.py
evangriffiths: My bad, I was copying this from another branch - it should be in the SellTokens description. Have moved! Hopefully it makes more sense in that context 😆 . I agree it's still confusing, but this can be removed when https://github.com/gnosis/prediction-market-agent-tooling/issues/195 is done

> I think phrasing it in terms of outcome tokens is easier for the agent

done

PR #186, opened 2024-05-15
Gather market predictions from existing correlated markets

gabrielfior:
- Think thoroughly agent now has access to correlated markets' probabilities (current_p_{yes,no}` as part of its final decision.
- Integrated Pinecone with PMA


prediction_market_agent/agents/think_thoroughly_agent/prompts.py
```diff
@@ -83,21 +83,20 @@
...
 
 SCENARIOS_WITH_PROBABILITIES: 
 {scenarios_with_probabilities}
 
-SCENARIO_TO_ASSESS: {scenario_to_assess}
-"""
+CORRELATED_MARKETS is also provided, which contains a list of scenarios that are semantically similar to SCENARIO_TO_ASSESS. The object RELATED_MARKETS contains a list of dictionaries, where each dictionary has properties such as question_title, as p_yes (probability that the scenario is true) and p_no (probability that the scenario is false).
+
+CORRELATED_MARKETS: 
+{correlated_markets}
```
gabrielfior, line 94: Main part - added correlated_markets to the prompt as reference.

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -32,8 +46,16 @@ class Scenarios(BaseModel):
...
+    pinecone_handler: t.Optional[PineconeHandler] = None
+    model: str
+
     def __init__(self, model: str) -> None:
         self.model = model
+        self.subgraph_handler = OmenSubgraphHandler()
+
+    def _init_pinecone(self) -> None:
+        if not self.pinecone_handler:
+            self.pinecone_handler = PineconeHandler()
```
gabrielfior, line 58: Loading it lazily in case the API key not present.

prediction_market_agent/agents/think_thoroughly_agent/deploy.py
```diff
@@ -17,9 +20,17 @@ def load(self) -> None:
     def answer_binary_market(self, market: AgentMarket) -> Answer | None:
         return self.agent.answer_binary_market(market.question)
 
+    def before(self, market_type: MarketType) -> None:
+        self.agent.update_markets()
```
gabrielfior, line 24: Note that `update_markets` is called, so that we constantly update the embeddings present on Pinecone every time `think-thoroughly` runs.
I was a bit conservative (checks for markets opened in prev 7 days) to account for failed runs. When uploading to Pinecone, it checks if market already there (by ID), and only uploads if market not already present.
Net effect on average is one extra subgraph call per run, so should not affect the run too much.

prediction_market_agent/agents/think_thoroughly_agent/deploy.py
```diff
@@ -17,9 +20,17 @@ def load(self) -> None:
...
+        super().before(market_type=market_type)
+
 
 if __name__ == "__main__":
     agent = DeployableThinkThoroughlyAgent(place_bet=False)
+    # We set the keys manually since it will be required by PMAT
+    if not os.environ.get("OPENAI_API_KEY"):
+        keys = APIKeys()
+        os.environ["OPENAI_API_KEY"] = keys.openai_api_key.get_secret_value()
```
kongzii, line 33: Why do you need to set this? Is `OPENAI_API_KEY` used somewhere directly from environment variable instead from APIKeys?

prediction_market_agent/agents/think_thoroughly_agent/pinecone_utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+from prediction_market_agent_tooling.markets.omen.data_models import OmenMarket
+
+from prediction_market_agent.agents.think_thoroughly_agent.models import (
+    PineconeMetadata,
+)
+
+
+def create_texts_from_omen_markets(markets: list[OmenMarket]) -> list[str]:
+    return [m.question_title for m in markets]
```
kongzii, line 11: `create_texts` is a little misleading, should it create some special texts or just extract the titles?

prediction_market_agent/agents/think_thoroughly_agent/pinecone_utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+def create_metadatas_from_omen_markets(
+    markets: list[OmenMarket],
+) -> list[dict[str, t.Any]]:
+    return [
+        PineconeMetadata.model_validate(
+            {"question_title": m.question_title, "market_address": m.id}
+        ).dict()
+        for m in markets
+    ]
```
kongzii, line 22: I don't understand this one, so it

1. Creates dictionary `{"question_title": m.question_title, "market_address": m.id}`
2. Creaets PineconeMetadata out of it
3. Converts it back to dictionary that should be same as in 1.?
4. Returns `dict[str, t.Any]` even though we know the exact structure?

Instead of this extra function, wdyt about doing this instead of calling  `create_metadatas_from_omen_markets`?

```
metadatas = [PineconeMetadata.from_omen_market(market).model_dump() for market in markets] 
```

(`dict` method is deprecated, `model_dump` should be used instead)

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
kongzii: Are they necessary correlated? Similar in the question for sure, but they don't have to be correlated. Or are you planning to do some filtering only for corelated ones?

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
kongzii: Why not just do `self.pinecone_handler = PineconeHandler()` inside of `__init__` ?

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -201,6 +256,7 @@ def generate_final_decision(
                 ),
                 "number_of_scenarios": len(scenarios_with_probabilities),
                 "scenario_to_assess": question,
+                "correlated_markets": [r.dict() for r in correlated_markets],
```
kongzii, line 259: I think dicts were messing up the prompt a little, wdyt about doing it in a human-readable way as above?

Screenshot by Dropbox Capture

prediction_market_agent/db/pinecone_handler.py
```diff
@@ -0,0 +1,55 @@
...
+from prediction_market_agent.agents.think_thoroughly_agent.models import (
+    PineconeMetadata,
+)
+from prediction_market_agent.utils import APIKeys
+
+INDEX_NAME = "omen-markets"
+
+
+# ToDo - Move to PMAT
+class PineconeHandler:
```
kongzii, line 18: If `PineconeHandler` is our class, why do we need type-ignores above in the code?

pyproject.toml
```diff
@@ -58,6 +58,7 @@ mech-client = { git = "https://github.com/valory-xyz/mech-client.git", rev = "4a
 streamlit-extras = "^0.4.2"
 tweepy = "^4.14.0"
```
kongzii, line 60:
```suggestion
```

scripts/embed_question.py
```diff
@@ -0,0 +1,35 @@
...
+    PineconeMetadata,
+)
+from prediction_market_agent.db.pinecone_handler import PineconeHandler
+
+
+def main() -> None:
+    """Script for inserting all open markets into Pinecone (if not yet there)."""
+    sh = OmenSubgraphHandler()
+    open_markets = sh.get_omen_binary_markets_simple(
+        limit=sys.maxsize, filter_by=FilterBy.OPEN, sort_by=SortBy.NEWEST
```
kongzii, line 19: Hmm, good question. Should it be open markest only? Resolved ones should have correct answers.

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -75,6 +97,23 @@ def _build_llm(self) -> BaseChatModel:
         )
         return llm
 
+    def update_markets(self) -> None:
+        """We use the agent's run to add embeddings of new markets that don't exist yet in the
+        vector DB."""
+        created_after = utcnow() - datetime.timedelta(days=7)
+        markets = self.subgraph_handler.get_omen_binary_markets_simple(
+            limit=sys.maxsize,
+            filter_by=FilterBy.OPEN,
```
kongzii, line 106: Same quesion as in the script, shouldn't we include resolved ones as well?

prediction_market_agent/agents/think_thoroughly_agent/deploy.py
gabrielfior: Yes, somewhere inside PMAT there is an LLM call that reads it from the .env.

prediction_market_agent/db/pinecone_handler.py
gabrielfior: mypy has trouble realizing that self.init_pinecone() initializes `self.pinecone_handler`, thus should not be of type `PineconeHandler|None` anymore, but rather `PineconeHandler`.
Hence the type ignore.

prediction_market_agent/agents/think_thoroughly_agent/pinecone_utils.py
gabrielfior: removed

prediction_market_agent/agents/think_thoroughly_agent/pinecone_utils.py
gabrielfior: Didn't know `dict()` was deprecated, thanks for the heads-up.
Sure, refactored.

scripts/embed_question.py
gabrielfior: Per our discussion, refactoring this to use all markets (scripts + agent usages).

prediction_market_agent/agents/think_thoroughly_agent/deploy.py
gabrielfior: https://github.com/gnosis/prediction-market-agent-tooling/blob/main/prediction_market_agent_tooling/tools/is_predictable.py#L47

gabrielfior: Creating a PR for that now

gabrielfior: Merged PR

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
```diff
@@ -83,21 +83,20 @@
...
-SCENARIOS_WITH_PROBABILITIES is provided, which contains a set of {number_of_scenarios} SCENARIOs.
-The object SCENARIOS_WITH_PROBABILITIES contains a list of tuple, where each tuple contains, as its
-first element, an scenario, and as its second entry, a probability dictionary, having properties such
-as p_yes (probability that the scenario is true), p_no (probability that the scenario is false),
-confidence (confidence level in the values of p_yes and p_no).
-You should determine the probability of the SCENARIO SCENARIO_TO_ASSESS being true, 
-considering the probabilities of the other related SCENARIOs.
+FINAL_DECISION_PROMPT = """Your task is to determine the probability of a SCENARIO being answered 'Yes' or 'No'. SCENARIOS_WITH_PROBABILITIES is provided, which contains a set of {number_of_scenarios} SCENARIOs. The object SCENARIOS_WITH_PROBABILITIES contains a list of tuple, where each tuple contains, as its first element, an scenario, and as its second entry, a probability dictionary, having properties such as p_yes (probability that the scenario is true), p_no (probability that the scenario is false), confidence (confidence level in the values of p_yes and p_no).
```
kongzii, line 86: Any reason to change form from bullet list to comma separated? I think that the first one was easier to read.

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -209,6 +227,21 @@ def generate_prediction_for_one_outcome(
             )
             return None
 
+    def get_correlated_markets(self, question: str) -> list[CorrelatedMarketInput]:
+        nearest_questions = self.pinecone_handler.find_nearest_questions(
+            5, text=question
```
kongzii, line 232: Did you play with this? Do you have some examples of what markets it fetches, please?

I'm thinking that if there are no similar markets, this way, it always returns 5 of them, even if they aren't similar at all. Maybe there should be some similarity threshold.

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
```diff
@@ -83,21 +83,20 @@
...
-You should determine the probability of the SCENARIO SCENARIO_TO_ASSESS being true, 
-considering the probabilities of the other related SCENARIOs.
+FINAL_DECISION_PROMPT = """Your task is to determine the probability of a SCENARIO being answered 'Yes' or 'No'. SCENARIOS_WITH_PROBABILITIES is provided, which contains a set of {number_of_scenarios} SCENARIOs. The object SCENARIOS_WITH_PROBABILITIES contains a list of tuple, where each tuple contains, as its first element, an scenario, and as its second entry, a probability dictionary, having properties such as p_yes (probability that the scenario is true), p_no (probability that the scenario is false), confidence (confidence level in the values of p_yes and p_no). 
 
 SCENARIOS_WITH_PROBABILITIES: 
 {scenarios_with_probabilities}
 
-SCENARIO_TO_ASSESS: {scenario_to_assess}
-"""
+CORRELATED_MARKETS is also provided, which contains a list of scenarios that are semantically similar to SCENARIO_TO_ASSESS. The object RELATED_MARKETS contains a list of dictionaries, where each dictionary has properties such as question_title, as p_yes (probability that the scenario is true) and p_no (probability that the scenario is false).
```
kongzii, line 91: >  contains a list of dictionaries, where each dictionary 

This is no longer true, right?

prediction_market_agent/db/pinecone_handler.py
```diff
@@ -0,0 +1,95 @@
...
+)
+from prediction_market_agent.utils import APIKeys
+
+INDEX_NAME = "omen-markets"
+T = t.TypeVar("T")
+
+
+class PineconeHandler:
+    def __init__(self) -> None:
+        k = APIKeys()
```
kongzii, line 28:
```suggestion
        keys = APIKeys()
```

Single-char variable names should almost never be used please 🙏

kongzii (approved):

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
```diff
@@ -107,8 +106,4 @@
 - "p_no": Probability that the sentence outcome will be `No`. Ranging from 0 (lowest probability) to 1 (maximum probability).
 - "confidence": Indicating the confidence in the estimated probabilities you provided ranging from 0 (lowest confidence) to 1 (maximum confidence). Confidence can be calculated based on the quality and quantity of data used for the estimation.
 
-Do not surround the output object with escape quotes, line breaks nor '''. 
-Do not output any other text, only the JSON object.
-
-Ensure p_yes + p_no equals 1.
-"""
+Do not surround the output object with escape quotes, line breaks nor '''. Do not output any other text, only the JSON object. Ensure p_yes + p_no equals 1."""
```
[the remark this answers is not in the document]
kongzii: Similar question here, why one-line everything?

scripts/embed_question.py
```diff
@@ -0,0 +1,12 @@
+import typer
+
+from prediction_market_agent.db.pinecone_handler import PineconeHandler
+
+
+def main() -> None:
+    """Script for inserting all open markets into Pinecone (if not yet there)."""
```
kongzii, line 7:
```suggestion
    """Script for inserting all markets into Pinecone (if not yet there)."""
```

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
gabrielfior: This wasn't changed in this PR.

gabrielfior: Motivation was to save tokens - I recall you raised a comment about it in a previous PR?
Not a problem to refactor though - are we prioritizing readability or token usage?

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
gabrielfior: Good point - that's true, not-so-correlated markets would be returned in this scenario.
I will add a score threshold here (already returned by Pinecone).

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
gabrielfior: Updating

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
gabrielfior: Implemented a filter with default `threshold=0.7` (rule-of-thumb value from a post), I think that already helps

kongzii: Are we getting any relevant markets from Omen with this?

prediction_market_agent/db/pinecone_handler.py
```diff
@@ -0,0 +1,112 @@
...
+                    texts=text_chunk, metadatas=metadata_chunk
+                )
+
+    def find_nearest_questions_with_threshold(
+        self, limit: int, text: str, threshold: float = 0.7
+    ) -> list[PineconeMetadata]:
+        # Note that pagination is not implemented in the Pinecone client.
+        # Hence we set a large limit and hope we get enough results that satisfy the threshold.
+        documents_and_scores = self.vectorstore.similarity_search_with_score(
+            query=text, k=int(limit * 5)
```
kongzii, line 102: why limit * 5?

kongzii: nevermind, noticed the comment now 😄

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
gabrielfior: We are fetching relevant markets - for ex, there are 2 markets related to Panama's election that can be cross-referenced.

kongzii (approved):

PR #190, opened 2024-05-16
Change LongTermMemory.search to take datetime filtering args

evangriffiths:
Fixes https://github.com/gnosis/prediction-market-agent/issues/185

### DIFF
--- a/prediction_market_agent/agents/microchain_agent/memory.py
+++ b/prediction_market_agent/agents/microchain_agent/memory.py
@@ -1,4 +1,5 @@
 # inspired by crewAI's LongTermMemory (https://github.com/joaomdmoura/crewAI/blob/main/src/crewai/memory/long_term/long_term_memory.py)
+from datetime import datetime
 from typing import Any, Dict, Sequence
 
 from prediction_market_agent.db.db_storage import DBStorage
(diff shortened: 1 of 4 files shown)


prediction_market_agent/agents/microchain_agent/memory.py
```diff
@@ -18,5 +19,13 @@ def save_history(self, history: list[Dict[str, Any]]) -> None:
             history=history,
         )
 
-    def search(self, latest_n: int = 5) -> Sequence[LongTermMemories]:
-        return self.storage.load(self.task_description, latest_n)
+    def search(
+        self,
+        from_: datetime | None = None,
+        to: datetime | None = None,
```
gabrielfior, line 25: Pedantic correction - we should have either `to_` and `from_` or both without underline.

gabrielfior (approved):

prediction_market_agent/agents/microchain_agent/memory.py
evangriffiths: we can't have `from` (without the underscore) because it's a python keyword. Adding an underscore to the end of keywords to use them as variables is a convention (according to PEP8 - see here https://peps.python.org/pep-0008/#descriptive-naming-styles):

> single_trailing_underscore_: used by convention to avoid conflicts with Python keyword, e.g. :
`tkinter.Toplevel(master, class_='ClassName')`

PR #201, opened 2024-05-17
Various improvements and fixes for microchain agent

evangriffiths:
UI improvements:
- When the user hits enter after adding some reasoning, the agent executes immediatedly (saves an extra click on Run for the user)
- Remove misc functions `Sum` and `Product`. They aren't much use atm.
- Increase the verbosity of the redeem function. Before it returned None.

Fixes:
- `get_balance` should show both xdai and wxdai. Previously it was showing just xdai. If the agent placed a bet, it could show the balance not changing at all!
- Include all agent functions in dropdown list (see pic)

Screenshot 2024-05-17 at 15 44 12

### DIFF
--- a/prediction_market_agent/agents/microchain_agent/utils.py
+++ b/prediction_market_agent/agents/microchain_agent/utils.py
@@ -54,9 +54,10 @@ def get_binary_markets(market_type: MarketType) -> list[AgentMarket]:
 def get_balance(market_type: MarketType) -> BetAmount:
     currency = market_type.market_class.currency
     if market_type == MarketType.OMEN:
-        # We focus solely on xDAI balance for now to avoid the agent having to wrap/unwrap xDAI.
+        balances = get_balances(APIKeys().bet_from_address)
+        total_balance = balances.xdai + balances.wxdai
         return BetAmount(
-            amount=get_balances(APIKeys().bet_from_address).xdai,
+            amount=total_balance,
             currency=currency,
         )
     else:
--- a/prediction_market_agent/agents/microchain_agent/app.py
+++ b/prediction_market_agent/agents/microchain_agent/app.py
@@ -20,8 +20,10 @@ from prediction_market_agent_tooling.tools.costs import openai_costs
 from prediction_market_agent_tooling.tools.utils import check_not_none
 from streamlit_extras.bottom_container import bottom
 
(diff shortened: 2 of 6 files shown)


prediction_market_agent/agents/microchain_agent/utils.py
```diff
@@ -54,9 +54,10 @@ def get_binary_markets(market_type: MarketType) -> list[AgentMarket]:
...
     currency = market_type.market_class.currency
     if market_type == MarketType.OMEN:
-        # We focus solely on xDAI balance for now to avoid the agent having to wrap/unwrap xDAI.
+        balances = get_balances(APIKeys().bet_from_address)
+        total_balance = balances.xdai + balances.wxdai
         return BetAmount(
-            amount=get_balances(APIKeys().bet_from_address).xdai,
+            amount=total_balance,
             currency=currency,
         )
```
gabrielfior, line 62: I find this a bit dangerous - indeed xDAI and wxDAI have the same value, but let's say the user has 10 xDAI and 20 wxDAI - he sees this balance and now wants to bet 30 units on a given market. Do we auto-trigger a conversion between them to allow for such a trade?
Another option would be to have 2 balances for trades (less nice in my opinion).

kongzii: Good point, @coderabbitai please create an issue to change all functions where `auto_deposit` is in the argument, to instead of depositing everything from xDai, deposit only the remainder of `available_wxdai - needed_wxdai`, for example here prediction_market_agent_tooling/markets/omen/omen.py

kongzii (approved):

prediction_market_agent/agents/microchain_agent/utils.py
evangriffiths: > Do we auto-trigger a conversion between them to allow for such a trade?

Yes, we do! See the `auto_deposit` arg https://github.com/gnosis/prediction-market-agent-tooling/blob/1e497fff9f2b53e4e3e1beb5dda08b4d49da881b/prediction_market_agent_tooling/markets/omen/omen.py#L139 (as @kongzii mentions).

So it's safe, but yes https://github.com/gnosis/prediction-market-agent/issues/206 would be a good optimization

PR #202, opened 2024-05-17
Tweaks for microchain agent deployment

evangriffiths:


prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -2,20 +2,27 @@
 from prediction_market_agent_tooling.deploy.agent import DeployableAgent
 from prediction_market_agent_tooling.markets.markets import MarketType
 
+from prediction_market_agent.agents.microchain_agent.memory import LongTermMemory
 from prediction_market_agent.agents.microchain_agent.microchain_agent import build_agent
 
 
 class DeployableMicrochainAgent(DeployableAgent):
-    model = "gpt-4-1106-preview"
+    model = "gpt-4o-2024-05-13"
```
kongzii, line 10: Are you sure? seems like `o` is generally laughed at 😄

evangriffiths: i'm a fanboy so i won't listen to the haters

PR #208, opened 2024-05-20
149 add long term memory to think thoroughly agent

### LINKED ISSUE #149: Add long term memory to think thoroughly agent
Currently, the microchain general agent has long term memory (i.e. it writes to the DB).
We should add this functionality to the `think-thoroughly` agent. This will allow the Farcaster agent to have access to its reasoning and thus produce better tweets/casts.

gabrielfior:
Closes #149


prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -78,8 +82,11 @@ async def _arun(
 
 
 class CrewAIAgentSubquestions:
-    def __init__(self, model: str) -> None:
+    def __init__(self, model: str, memory: bool = True) -> None:
         self.model = model
+        self.memory = memory
+        if self.memory:
+            self._long_term_memory = LongTermMemory("think-thoroughly-agent")
```
gabrielfior, line 89: Note that, as discussed with @evangriffiths , we won't filter by question (no such clear division on the microchain agent, which also shares memory), but rather filter only on task-datetime.

kongzii: Seems like long_term_memory can be undefined if memory=false, but later in the methods it is accessed unconditionally

gabrielfior: Great catch! Fixed that

gabrielfior:
Changes from this PR were included in a different PR (#214 ).
Closing this.

PR #219, opened 2024-05-30
Add monitor script

kongzii:
Add monitor app to PMA instead of PMAT, because in PMA we already have dockerfile, building and pushing of it. So we don't have to duplicate that logic.


gabrielfior:
It would be nice (note to myself included) to add a screenshot of the final Streamlit app, it helps the reviewer picture the PR quicker (before starting the app).

kongzii:
@gabrielfior This is an old app, already deployed via streamlit cloud https://prediction-market-agent-tooling-monitor.streamlit.app. It just imports it from PMAT, no new stuff in this PR.

PR #220, opened 2024-05-30
Deploying script

kongzii:


gabrielfior (changes requested):
Can we also add instructions to the README please?

kongzii:
@gabrielfior Should it be in the README if it's an open source repository and this is specific only to us?

gabrielfior:
> @gabrielfior Should it be in the README if it's an open source repository and this is specific only to us?

Then the script should not be in the open source repo, but rather in a different repo (e.g. `deployments-gcp`).
But sure, OK for me for now.

gabrielfior (approved):

PR #222, opened 2024-05-31
More generalistic general agent thanks to learning

kongzii:


prediction_market_agent/agents/microchain_agent/agent_functions.py
```diff
@@ -0,0 +1,55 @@
...
+
+from prediction_market_agent.agents.microchain_agent.prompts import (
+    NON_UPDATABLE_DIVIDOR,
+    SYSTEM_PROMPT,
+)
+
+
+class AgentAction(Function):
+    def __init__(self, agent: Agent) -> None:
+        self.agent = agent
```
kongzii, line 13: These functions allow to change the agent itself

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -171,19 +183,37 @@ def get_function_bullet_point_list(model: str) -> str:
...
+                "gpt-3.5-turbo-0125",
+                "gpt-4o-2024-05-13",
+            ],
             index=0,
         )
         if model is None:
             st.error("Please select a model.")
+        st.session_state.system_prompt_select = st.selectbox(
+            "Initial system prompt",
+            [p.value for p in SystemPromptChoice],
```
kongzii, line 197: See the `SystemPromptChoice` later on but the idea is, that there are two choices:

1. "plain" or "from scratch" choice, where the agent is supposed to learn everything
2. "trader" choice, where the system prompt is the same thing as was before, to start trading right away

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -23,6 +26,7 @@ def run(self, market_type: MarketType) -> None:
         agent: Agent = build_agent(
             market_type=market_type,
             model=self.model,
+            system_prompt=TRADING_AGENT_SYSTEM_PROMPT,  # Use pre-learned system prompt until the prompt-fetching from DB is implemented.
```
kongzii, line 29: After the agent updates its system prompt, it needs to be saved in the DB and then automatically fetches. This will bring some complication so I'd leave that for the next PR.

Right now, DeployableAgent is using `trader` prompt right away (instead of always learning from scratch) and the learning prompt is used only in the streamlit demo to showcase it.

prediction_market_agent/agents/microchain_agent/learning_functions.py
```diff
@@ -0,0 +1,70 @@
+import typing as t
+
+from microchain import Function
+
+
+class LearningFunction(Function):
+    title: str
+    knowledge: str
```
kongzii, line 8: Various functions that agent must use to learn about how to work.

prediction_market_agent/agents/microchain_agent/market_functions.py
```diff
@@ -319,38 +312,6 @@ def __call__(self) -> list[str]:
         return [str(position) for position in positions]
 
 
-class RememberPastLearnings(Function):
```
kongzii, line 322: Just moved to another file as this isn't a market function

kongzii:
Streamlit demo updates:

Screenshot by Dropbox Capture

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -3,33 +3,48 @@
...
-SYSTEM_PROMPT = """
-Act as a trader agent in prediction markets to maximise your profit.
 
-Research markets, buy tokens you consider undervalued, and sell tokens that you
-hold and consider overvalued.
+class CustomizedAgent(Agent):
+    """
+    Subclass microchain's agent for any customizations that are needed right away.
+    Anything generally useful should be moved upstream to the microchain repository.
+    """
```
gabrielfior, line 34: I don't follow - is `Agent` intended to be an abstract class? if yes, we should inherit from ABC to make it explicit.

Another question - we only have 1 microchain agent, why do you think it's a good idea to use inheritance here?

prediction_market_agent/agents/microchain_agent/deploy.py
gabrielfior: Sorry, also don't follow here - my understanding was this
-> General agent performs a bunch of actions, incl. learning but hopefully trading. It stores its actions in the DB, as well as the open positions are also queryable (e.g. via `getUserPositions`).
-> The prompt was continuously updated during the run (appending history), now the agent has the freedom to update it (this PR). OK.
-> In my head, once the run is over, the prompt is lost - i.e. a new run must start with the original prompt. However (this is the important part), by calling `getPastLearnings`, the agent can see what he learned previously (from the activities in the current table) and hence accelerate its learning.

This is all to say - I don't think we should store the prompts in the DB. What do you think?

kongzii: We need to. The system prompt (how the agent should behave) and the history of user prompts (actions that the agent took) are two different concepts.

Another thing I wanted to propose today is renaming `RememberPastLearnings` to something like `RememberPastActions`, because even in the description of that class there is `Use this function to fetch information about the actions`.

kongzii: As an example, let's say the agent starts with the system prompt `Learn, survive and evolve` and it evolved to a system prompt:

```
Act as a trader agent in prediction markets to maximise your profit.

Research markets, buy tokens you consider undervalued, and sell tokens that you
hold and consider overvalued.
```

(this is a copy-paste of the original system prompt, but good for this example, it could evolve to something similar)

We don't want him next day to start again as `Learn, evolve survive`, it already learned trading and stuff, so we want the agent to continue, not always start from 0.

prediction_market_agent/agents/microchain_agent/microchain_agent.py
kongzii: PR opened here https://github.com/galatolofederico/microchain/pull/11

When it's merged I'll remove `build_initial_messages` in this `CustomizedAgent` class and we can either keep it  plain or remove completely -- no preference on my side atm.

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -41,47 +56,49 @@ def build_agent_functions(
...
 
+    functions.extend([f() for f in LEARNING_FUNCTIONS])
+    functions.extend([f(agent=agent) for f in AGENT_FUNCTIONS])
     functions.extend([f(market_type=market_type) for f in MARKET_FUNCTIONS])
     if market_type == MarketType.OMEN:
         functions.extend([f() for f in OMEN_FUNCTIONS])
     if long_term_memory:
         functions.append(
-            RememberPastLearnings(long_term_memory=long_term_memory, model=model)
+            RememberPastActions(long_term_memory=long_term_memory, model=model)
```
kongzii, line 66: It's the same thing, but renamed for clarity, as mentioned in the other comment.

prediction_market_agent/agents/microchain_agent/agent_functions.py
evangriffiths: Something about this doesn't feel right. Intuitively I would say that an agent should have a Function. But not the other way around. Can't think of another way of doing this with microchain though...

evangriffiths: > Intuitively I would say that an agent should have a Function. But not the other way around

That's how crewai/autogen have it:

```
        llm_config["tools"] = [google_search_tool.schema, web_scaping_tool.schema]
        self.assistant = GPTAssistantAgent(
            name="assistant",
            instructions="...",
            llm_config=llm_config,
        )
```
so I wonder if this kind of behaviour is possible with these frameworks.

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -204,12 +235,25 @@ def get_function_bullet_point_list(model: str) -> str:
...
-    st.markdown(get_function_bullet_point_list(model=model))
+    st.markdown("It is equipped with the following tools:")
+    if agent_is_initialized():
+        st.markdown(
+            get_function_bullet_point_list(agent=st.session_state.agent, model=model)
+        )
+    else:
+        st.markdown("The agent is not initialized yet.")
+
+with st.expander("Agent's current system prompt"):
```
evangriffiths, line 246: I haven't done a very good job of it in this app.py, but the intention is to have a separation between layout (the top part, where everything that's not static is just a container) and execution (the bottom part under `# Execution`).

The idea being that we don't always want things to run in the same order that they are laid out in. And also it should be easy-ish from just reading the code what the layout will look like, and what order things will run in. These new bits are dynamic, so I guess there should be a placeholder here and the markdown should be generated further down

evangriffiths: @gabrielfior noticed that as it currently is, the prompt is updated in the UI with a 1-iteration lag, because it gets written before the agent is run (and potentially does calls 'update system prompt').

I know you're looking at this in https://github.com/gnosis/prediction-market-agent/pull/238. What do you think about this suggested approach?

evangriffiths (approved):
Approving so it can be merged now. Would be great to still get responses on the feedback and have follow-on PRs if necessary

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -204,12 +235,25 @@ def get_function_bullet_point_list(model: str) -> str:
...
+    if agent_is_initialized():
+        st.markdown(st.session_state.agent.system_prompt)
```
evangriffiths, line 248: If we've already got the 'info' dropdown box that states the tools the agent has, then I think it would look nicer if this only contained the system prompt above `NON_UPDATABLE_DIVIDOR`, so we don't have duplicate-ish info

prediction_market_agent/agents/microchain_agent/app.py
gabrielfior: I will create a separate PR for solving this 1-iteration lag (spoiler alert - it involves adding callbacks to streamlit elements like buttons).
Regarding separating `Execution` and `Layout`, also fine and fits nicely with my changes from above.

prediction_market_agent/agents/microchain_agent/agent_functions.py
kongzii: Yep, it feels odd; a function like this can't even come from the marketplace. 

So if you have a better idea how to implement this I'm open to it 😄 Maybe the agent class could have something like `get_my_functions(self) ->? list[AgentFunction]` method?

prediction_market_agent/agents/microchain_agent/agent_functions.py
```diff
@@ -0,0 +1,90 @@
...
+    @property
+    def description(self) -> str:
+        return "Use this function to get your current system prompt."
+
+    @property
+    def example_args(self) -> list[str]:
+        return []
+
+    def __call__(self) -> str:
+        return str(self.agent.system_prompt).split(NON_UPDATABLE_DIVIDOR)[0].strip()
```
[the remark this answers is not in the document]
kongzii: Sounds better, @coderabbitai please create an issue for this.

prediction_market_agent/agents/microchain_agent/agent_functions.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def __call__(self) -> str:
+        return (
+            str(self.agent.bootstrap[0])
+            if self.agent.bootstrap
+            else "No bootstrap available."
+        )
+
+
+class UpdateMyBootstrap(AgentAction):
```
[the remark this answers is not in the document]
kongzii: Sounds like it should work, @coderabbitai please create an issue for this, and mention that it needs to be tested.

prediction_market_agent/agents/microchain_agent/app.py
kongzii: Sounds good, @coderabbitai please create an issue for this

prediction_market_agent/agents/microchain_agent/agent_functions.py
evangriffiths: > Maybe the agent class could have something like get_my_functions(self) -> list[AgentFunction] method

Yes that sounds like a good idea. Probably not high priority, so I'll just make a task to think about it more for now https://github.com/gnosis/prediction-market-agent/issues/253

PR #238, opened 2024-06-11
230 saving system prompt to db + streamlit fixes

gabrielfior:
-> Prompts are loaded from DB (if toggle activated)
-> Prompts are stored into DB
-> Fixed lagging of user prompt display


prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -246,24 +282,56 @@ def get_function_bullet_point_list(agent: Agent, model: str) -> str:
...
+        iterations=int(iterations),
+        model=model,
+    )
+    save_last_turn_history_to_memory(st.session_state.agent)
+
+
 # Placeholder for the agent's history
-history_container = st.container()
+with st.container() as history_container:
+    if "agent" in st.session_state:
```
evangriffiths, line 316:
```suggestion
    if agent_is_initialized():
```

tests/db/test_db_storage.py
```diff
@@ -1,13 +1,17 @@
...
 from prediction_market_agent_tooling.tools.utils import utcnow
+from sqlalchemy import Table
+from sqlmodel import SQLModel
 
 from prediction_market_agent.db.db_storage import DBStorage
+from prediction_market_agent.db.models import LongTermMemories, Prompt
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="function")
```
gabrielfior, line 14: Changed scope to `function` to avoid handling setup and tear down, rather having a fresh DB for every test.

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -80,8 +87,22 @@ def build_agent(
         engine.register(f)
 
     agent.max_tries = 3
+
+    # We restore the prompt from a historical session.
+    if prompt_handler:
+        historical_prompt = prompt_handler.fetch_latest_prompt()
+        system_prompt = historical_prompt.prompt
```
gabrielfior, line 94: This is responsible for overwriting `system_prompt` with the latest value from the DB.

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -255,15 +267,40 @@ def get_function_bullet_point_list(agent: Agent, model: str) -> str:
...
+    # st.session_state.system_prompt_test = random.randint(1000, 2000)
+    maybe_initialize_agent(model, system_prompt, bootstrap)
+
+    run_agent(
+        agent=st.session_state.agent,
+        iterations=int(iterations),
+        model=model,
+    )
+    save_last_turn_history_to_memory(st.session_state.agent)
+
```
gabrielfior, line 281: Main change in the streamlit app - replacing the main file flow with several `if` statements with functions that are called when button is clicked (hence the `on_click...` pattern).
This fixed the lag when displaying the system_prompt inside the history container.

gabrielfior:
@evangriffiths here's how the app looks like now (see toggle on bottom left)

image

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -165,8 +174,11 @@ def get_function_bullet_point_list(agent: Agent, model: str) -> str:
 with st.sidebar:
     streamlit_login()
 check_required_api_keys(["OPENAI_API_KEY", "BET_FROM_PRIVATE_KEY"])
-keys = APIKeys()
-maybe_initialize_long_term_memory()
+
+
+def maybe_init() -> None:
```
evangriffiths, line 179: Looks like this wrapper isn't useful anymore.

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -80,8 +83,22 @@ def build_agent(
         engine.register(f)
 
     agent.max_tries = 3
+
+    # We restore the prompt from a historical session.
+    if prompt_handler:
+        if historical_prompt := prompt_handler.fetch_latest_prompt():
+            system_prompt = historical_prompt.prompt
+
     print(system_prompt)
```
evangriffiths, line 92: While we're here, delete this print

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -80,8 +83,22 @@ def build_agent(
...
-    agent.system_prompt = system_prompt.format(engine_help=engine.help)
+    # if {engine_help} not in prompt, we expect the functions to have been already loaded,
+    # thus no need to load them again. Otherwise we can simply not use the historical prompt.
+    agent.system_prompt = system_prompt
+    if "{engine_help}" not in system_prompt:
```
evangriffiths, line 96: This feels quite hacky. This is something I didn't pick up in https://github.com/gnosis/prediction-market-agent/pull/222, but I think `build_agent` shouldn't take a system prompt template as an arg, it should take the fully-formatted system prompt. Then we don't need to make any assumptions about what the prompt contains here.

The way I'd do that is by moving the engine initialization into a separate function, `build_agent_engine`, and then `build_agent` takes an engine arg, and `system_prompt` is the actual prompt. wdyt?

prediction_market_agent/db/db_storage.py
```diff
@@ -6,10 +6,15 @@
...
-from prediction_market_agent.db.models import LongTermMemories
+from prediction_market_agent.db.models import (
+    PROMPT_DEFAULT_SESSION_IDENTIFIER,
+    LongTermMemories,
+    Prompt,
+)
 from prediction_market_agent.utils import DBKeys
 
 
+# ToDo - Split this into base class and table specific classes
```
evangriffiths, line 17: Same as above

prediction_market_agent/agents/microchain_agent/prompt_handler.py
```diff
@@ -0,0 +1,34 @@
+from loguru import logger
+from prediction_market_agent_tooling.tools.utils import utcnow
+
+from prediction_market_agent.db.db_storage import DBStorage
+from prediction_market_agent.db.models import Prompt
+
+
+# ToDo - Unify PromptHandler, db_storage and LongTermMemory into 2 classes, one per table.
```
evangriffiths, line 8: Do you want to do that as part of this PR or a follow-up one? If follow-up, could you make an issue and tag it here?

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -255,15 +269,39 @@ def get_function_bullet_point_list(agent: Agent, model: str) -> str:
     else:
         st.markdown("The agent is not initialized yet.")
 
+
+def on_click_run_agent() -> None:
```
evangriffiths, line 273: Could you move this to the top section with the rest of the functions, just for readability of the streamlit app code?

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -275,39 +313,6 @@ def get_function_bullet_point_list(agent: Agent, model: str) -> str:
         with col3:
             st.caption(" \- OR -")
         with col4:
-            user_reasoning = st.chat_input("Add reasoning...")
-
-# Execution
```
evangriffiths, line 280: I'm confused!

Here: https://github.com/gnosis/prediction-market-agent/pull/222#discussion_r1635226431 you say

> Regarding separating Execution and Layout, also fine and fits nicely with my changes from above.

but now execution is all inline with layout!!

evangriffiths (commented):
Nice! PromptHandler implementation fits nicely with existing memory logic. Have left a few inline comments/requested changes.

prediction_market_agent/agents/microchain_agent/app.py
gabrielfior: removed

prediction_market_agent/agents/microchain_agent/prompt_handler.py
gabrielfior: There is a separate issue for that, see https://github.com/gnosis/prediction-market-agent/issues/242

prediction_market_agent/agents/microchain_agent/app.py
gabrielfior: Please see latest state of app.py -> https://github.com/gnosis/prediction-market-agent/blob/1a3e9421191e461c696b05c1117c06e4bd954bb1/prediction_market_agent/agents/microchain_agent/app.py#L188

What I meant was: from line 188, we only have layout elements that reference st.state_layout.<VARIABLE>. All the functions are above and do not get called.
Are you ok with this current structure?

prediction_market_agent/agents/microchain_agent/microchain_agent.py
gabrielfior: Agree it's a bit hacky and like the overall idea, however not sure how to handle the fact that this assignment (L95)
```
agent.system_prompt = system_prompt
```
is needed for agent initialization. Hence we need the system_prompt here still.

My counter-proposal would be to move the hacky bit to a separate function (`maybe_append_functions_into_prompt` or similar), and that gets called during agent initialization.
Bear in mind it's nothing magical, it simply adds the function definitions to the prompt being initialized in case it is still not there. Wdyt?

gabrielfior: See https://github.com/gnosis/prediction-market-agent/commit/d2d892136b033d71255e18ba62b2860bb527fc1d#diff-59ec0cef505f2b78726896b047967f06a472aeb8b7bb5ec33baab7e43fff720aR57 for the current suggestion

evangriffiths: > however not sure how to handle the fact that this assignment ...

Not sure I understand. That could just stay where it is?

So before:
```
def build_agent(
    system_prompt: str,
    ....
):
    engine = Engine()
    agent = Agent(engine, ...)
    agent.system_prompt = system_prompt.format(engine_help=engine.help)
```

After:

```
engine = ...

def build_agent(
    system_prompt: str,
    engine: Engine,
    ....
):
    agent = Agent(engine)
    agent.system_prompt = system_prompt
```

But my point is that sequence of operations should be:

- Build the engine
- Build the system prompt, either by formatting `TRADING_AGENT_SYSTEM_PROMPT` or by fetching a prompt from the db (not by looking for a particular string in the prompt)
- Build the agent.

With `maybe_append_function_definitions_into_prompt` the hackyness is just moved somewhere else.

prediction_market_agent/agents/microchain_agent/prompt_handler.py
evangriffiths: Could you link that PR here so it doesn't get lost?

prediction_market_agent/agents/microchain_agent/app.py
evangriffiths: What's `st.state_layout`?

For the balance lag issue, here's what I had in mind for a fix https://github.com/gnosis/prediction-market-agent/pull/247. Can we do something similar here for the system prompt?

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -48,20 +48,38 @@
...
                 agent.run(iterations=1, resume=True)
         st.session_state.running_cost += costs.cost
 
 
-def execute_reasoning(agent: Agent, reasoning: str, model: str) -> None:
-    with openai_costs(model) as costs:
-        agent.execute_command(f'Reasoning("{reasoning}")')
-        display_new_history_callback(agent)  # Run manually after `execute_command`
+def on_click_run_agent() -> None:
+    maybe_initialize_agent(system_prompt, bootstrap)
```
kongzii, line 61: Shouldn't this be in ` run_agent` function as well?

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -48,20 +48,38 @@
...
+    run_agent(
+        agent=st.session_state.agent,
+        iterations=int(iterations),
```
kongzii, line 64:
```suggestion
        iterations=iterations,
```

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -48,20 +48,38 @@
...
+
+
+def on_click_user_reasoning() -> None:
+    maybe_initialize_agent(system_prompt, bootstrap)
+
+    with openai_costs(st.session_state.model) as costs:
+        st.session_state.agent.execute_command(f'Reasoning("{user_reasoning}")')
         st.session_state.running_cost += costs.cost
 
+    save_last_turn_history_to_memory(st.session_state.agent)
```
kongzii, line 76: Above, this is call after running the agent, here it is called before running the agent. What's the correct way?

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -103,13 +122,9 @@
...
 
 
-def long_term_memory_is_initialized() -> bool:
-    return "long_term_memory" in st.session_state
-
-
 def maybe_initialize_long_term_memory() -> None:
     # Initialize the db storage
-    if not long_term_memory_is_initialized():
+    if not "long_term_memory" in st.session_state:
```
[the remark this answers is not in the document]
kongzii: Rabbit is correct, however, why did you remove `long_term_memory_is_initialized` function?

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -176,42 +192,38 @@ def get_function_bullet_point_list(agent: Agent, model: str) -> str:
...
+        key="system_prompt_select",
+        disabled=agent_is_initialized(),
+    )
+    st.toggle(
+        "Load historical prompt",
+        key="load_historical_prompt",
+        disabled=agent_is_initialized(),
+    )
+
+    # model = check_not_none(model)
```
kongzii, line 226:
```suggestion
```

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -255,15 +265,28 @@ def get_function_bullet_point_list(agent: Agent, model: str) -> str:
     else:
         st.markdown("The agent is not initialized yet.")
 
+
 # Placeholder for the agent's history
-history_container = st.container()
+with st.container() as history_container:
```
kongzii, line 270: If `container` is created and used and the same place, there is no need for the container, is there?

prediction_market_agent/agents/microchain_agent/prompt_handler.py
```diff
@@ -0,0 +1,34 @@
...
+        self,
+        session_identifier: str | None = None,
+        sqlalchemy_db_url: str | None = None,
+    ):
+        self.session_identifier = session_identifier
+        self.storage = DBStorage(sqlalchemy_db_url=sqlalchemy_db_url)
+        self.storage._initialize_db()
+
+    def save_prompt(self, prompt: str) -> None:
+        """Save item to storage. Note that score allows many types for easier handling by agent."""
```
kongzii, line 20: What score?

prediction_market_agent/agents/microchain_agent/prompt_handler.py
```diff
@@ -0,0 +1,34 @@
...
+        if not prompt:
+            logger.info("No prompt to save")
+            return
```
kongzii, line 23: Why would there be an empty string?

prediction_market_agent/db/db_storage.py
```diff
@@ -24,27 +28,50 @@ def _initialize_db(self) -> None:
         """
 
         # trick for making models import mandatory - models must be imported for metadata.create_all to work
-        logger.debug(f"tables being added {LongTermMemories}")
+        logger.debug(f"tables being added {LongTermMemories} {Prompt}")
         SQLModel.metadata.create_all(self.engine)
 
-    def save_multiple(
+    def save_multiple(self, items: list[Any]) -> None:
```
kongzii, line 34: Are you sure that `Any`thing is savable in the DB?

prediction_market_agent/db/db_storage.py
```diff
@@ -24,27 +28,50 @@ def _initialize_db(self) -> None:
...
+        self.save_multiple(history_items)
+
+    def load_latest_prompt(
+        self, session_identifier: str | None = None
+    ) -> Prompt | None:
+        """
+        Queries the prompts table by session identifier with error handling
+        """
+        if not session_identifier:
+            session_identifier = PROMPT_DEFAULT_SESSION_IDENTIFIER
```
kongzii, line 64: Why not just make it a default argument in the function definition? But should there be a default value?

prediction_market_agent/db/models.py
```diff
@@ -11,3 +11,21 @@ class LongTermMemories(SQLModel, table=True):
...
+
+    __tablename__ = "prompts"
+    __table_args__ = {
+        "extend_existing": True
+    }  # required if initializing an existing table
+    id: Optional[int] = Field(default=None, primary_key=True)
+    prompt: str
+    # This allows for future distinction between user sessions, if prompts from a specific
+    # user (or app) should be persisted.
+    session_identifier: Optional[str] = Field(default=PROMPT_DEFAULT_SESSION_IDENTIFIER)
```
kongzii, line 30:
```suggestion
    session_identifier: str
```

This default value is specified on multiple places, that becomes error prone after a while

prediction_market_agent/agents/microchain_agent/app.py
evangriffiths: Here's the equivalent kind of change to fix the system prompt/bootstrap lag https://github.com/gnosis/prediction-market-agent/pull/254

prediction_market_agent/db/models.py
gabrielfior: Sure, agree with the reasoning of maintainability - added default values one step before writing/reading from DB instead of defining the default value on a DB level. Hope that's fine now.

prediction_market_agent/agents/microchain_agent/prompt_handler.py
gabrielfior: I saw occurances of this due to the prompt lagging that was fixed previously - hence added this safety check.

prediction_market_agent/agents/microchain_agent/prompt_handler.py
gabrielfior: removed

prediction_market_agent/agents/microchain_agent/app.py
gabrielfior: Refactored, resolved

prediction_market_agent/agents/microchain_agent/app.py
gabrielfior: @evangriffiths could you please help me understand the logic here? https://github.com/gnosis/prediction-market-agent/blob/db46c34c87aed6e6b1f7582a7771b363c2fb7a7f/prediction_market_agent/agents/microchain_agent/app.py#L268-L281
My questions:
-> Why do you call `agent.run` after `agent.execute_command`?  From the [Agent file](https://github.com/galatolofederico/microchain/blob/20320cded714f12a11e7f237671faec4174b55b1/microchain/engine/agent.py#L4), I understand that `agent.execute_command` and `agent.run` behave similarly, hence only one is needed.
-> If you agree with the above, I would suggest removing the subsequent call to `run_agent` after the `execute_reasoning` is called (line 275 approx).

prediction_market_agent/agents/microchain_agent/app.py
gabrielfior: Not needed anymore, other branch was used

prediction_market_agent/agents/microchain_agent/app.py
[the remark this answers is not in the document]
gabrielfior: Not relevant anymore

prediction_market_agent/agents/microchain_agent/app.py
gabrielfior: I merged `app.py` from main, which comes from the PR you mentioned. I believe this solves this issue.

gabrielfior:
Problem still occurs with the general agent when using the `just-born` prompt, the agent just calls functions wrongly - this is in the branch 230-...., see app.py (version from main)

image

@evangriffiths @kongzii

prediction_market_agent/agents/microchain_agent/prompt_handler.py
kongzii: So is no longer needed if the lag is fixed?

prediction_market_agent/db/db_storage.py
kongzii: In `test_save_prompt` you are building `Prompt` class and then saving it:

```
prompt_text = "abc"
prompt = Prompt(prompt=prompt_text, datetime_=utcnow())
db_storage_test.save_multiple([prompt]
```

But if this is typed as `Any`, it means that even `db_storage_test.save_multiple(["hello", 123, 42.0]` should work,  but that will error out, won't it?

kongzii:
> Problem still occurs with the general agent when using the `just-born` prompt, the agent just calls functions wrongly - this is in the branch 230-...., see app.py (version from main)

@gabrielfior What model are you using? It was working very nicely before with gpt-4-turbo.

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -52,6 +53,20 @@
...
+    """
+    Updates the system prompt of the agent based on the availability of engine help.
+    """
+    # If {engine_help} not in prompt, we expect the functions to have been already loaded,
+    # thus no need to load them again. Otherwise, we can simply not use the historical prompt.
+    if "{engine_help}" in system_prompt:
+        system_prompt.format(engine_help=agent.engine.help)
+    agent.engine.help_called = True
+    agent.system_prompt = system_prompt
+
```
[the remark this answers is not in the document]
evangriffiths: Coderabbit is correct!

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -182,37 +186,31 @@ def get_function_bullet_point_list(agent: Agent, model: str) -> str:
...
-            disabled=True,
-        )
-        st.selectbox(
-            "Initial memory",
-            [st.session_state.system_prompt_select],
-            index=0,
-            disabled=True,
-        )
-    model = check_not_none(model)
+    st.selectbox(
```
evangriffiths, line 189: These look like nice changes (though not related to saving system prompt, right?). Another issue the app has on main is that the disabling of these select boxes also has a lag - the disabling happens when you run the agent for the second time, instead of at the end of the first run. Do you know if it's possible to move the disabling to after the agent has been run, to fix this?

prediction_market_agent/agents/microchain_agent/app.py
evangriffiths: That was meant to be a UX improvement. Before the user did:
1. type in some reasoning, hit enter -> the agent executes the reasoning, adds it to its history, and this is displayed in the app
2. the user clicks on 'run agent'

Since I assumed (1) is always followed by (2), I just merged them for convenience.

evangriffiths:
> Problem still occurs with the general agent when using the just-born prompt, the agent just calls functions wrongly

I was testing on main a bunch yesterday with gpt4o and didn't see that problem. Looks like it might not have `engine.help` in its system prompt?

gabrielfior:
> you using? It was working very

I still see this using gpt4, see screenshot below. I will open a GH issue.

image

kongzii:
> I still see this using gpt4, see screenshot below. I will open a GH issue.

@gabrielfior  If it works for both me and Evan, it's likely bug coming from this PR isn't it?

prediction_market_agent/agents/microchain_agent/microchain_agent.py
gabrielfior: Created this issue for tracking this discussion. Let's make this part of a separate PR (agree with the reasoning though).

prediction_market_agent/agents/microchain_agent/app.py
gabrielfior: Nice, thanks. So I changed the button to `add reasoning and run the agent` for clarity.

prediction_market_agent/agents/microchain_agent/prompt_handler.py
gabrielfior: sure, removing

prediction_market_agent/db/db_storage.py
gabrielfior: It's typed as `Any` since it's a generic function for saving any items (hence the `save_multiply` name).
Do you have a suggestion about how to make this better?

kongzii: I see your reasoning now.

Typing should reflect how/if the code will work. `typing.Any` is literally anything, but afaik this function will fail if you try to save anything that isn't a SQLModel's subclass.

So 

```suggestion
    def save_multiple(self, items: list[SQLModel]) -> None:
```

looks correct. It can save any SQLModel's data.

prediction_market_agent/agents/microchain_agent/app.py
gabrielfior: Great question - I don't know the answer, but when I do this (using "key" and reading from `session_state`), it works like a charm:
```
st.toggle("on_off", key="on_off")
st.write(f"on-off {st.session_state.on_off}")
st.button("will be deactivated", disabled=st.session_state.on_off)
```
So my guess is that the check `agent in st.session_state` is not updating properly, whereas the update is more "synchronous" when setting keys via streamlit widgets.

gabrielfior:
> > I still see this using gpt4, see screenshot below. I will open a GH issue.
> 
> @gabrielfior If it works for both me and Evan, it's likely bug coming from this PR isn't it?
 
> > Looks like it might not have engine.help in its system prompt?
> 
> This sounds possible, no idea why it would try to call something as `Learn("survival strategies")` otherwise.

Weird behavior not seen anymore. Not really a bug as I didn't have to change anything. Closing the other issue as well.

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -108,12 +123,21 @@ def main(
...
     agent.run(iterations=iterations)
     # generator.print_usage() # Waiting for microchain release
     long_term_memory.save_history(agent.history)
+    editable_prompt = get_editable_prompt_from_agent(agent)
+    prompt_handler.save_prompt(editable_prompt)
+
+
+def get_editable_prompt_from_agent(agent: Agent) -> str:
+    # Split prompt into editable, non-editable
+    editable_prompt = str(agent.system_prompt).split(NON_UPDATABLE_DIVIDOR)[0].strip()
```
kongzii, line 139:
```suggestion
    editable_prompt =agent.system_prompt.split(NON_UPDATABLE_DIVIDOR)[0].strip()
```

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -80,8 +83,18 @@ def build_agent(
         engine.register(f)
 
     agent.max_tries = 3
-    print(system_prompt)
-    agent.system_prompt = system_prompt.format(engine_help=engine.help)
+
+    # Restore the prompt from a historical session, replacing the editable part with it.
+    if prompt_handler:
+        if historical_prompt := prompt_handler.fetch_latest_prompt():
```
kongzii, line 89: In PromptHandler's `save_prompt` function, it's accessing session identifier from `self`:

```
def save_prompt(self, prompt: str) -> None:
        """Save item to storage."""
        prompt_to_save = Prompt(
            prompt=prompt,
            datetime_=utcnow(),
            session_identifier=self.session_identifier
            if self.session_identifier
            else PROMPT_DEFAULT_SESSION_IDENTIFIER,
        )
        self.storage.save_multiple([prompt_to_save])
```

However, in the `fetch_latest_prompt` method, session identifier is expected as an argument:

```
def fetch_latest_prompt(
        self, session_identifier: str = PROMPT_DEFAULT_SESSION_IDENTIFIER
    ) -> Prompt | None:
        return self.storage.load_latest_prompt(session_identifier=session_identifier)
```

So here at this line, it always fetches prompt for the `PROMPT_DEFAULT_SESSION_IDENTIFIER`, instead of the actual session identifier in PromptHandler.

Btw that's why I suggested before that the session identifier shouldn't be a default argument because it's easy to miss a bug like this.

prediction_market_agent/agents/microchain_agent/microchain_agent.py
kongzii: Closing this one as microchain's system prompt isn't typed, so we need this weird str(...) :(

prediction_market_agent/agents/microchain_agent/microchain_agent.py
gabrielfior: In the comments below
https://github.com/gnosis/prediction-market-agent/pull/238#discussion_r1642432944
https://github.com/gnosis/prediction-market-agent/pull/238#discussion_r1642434410

you said
> Why not just make it a default argument in the function definition? But should there be a default value?

So I'm not 100% sure we both agree on the implementation. Happy to refactor this down the line and discuss more - for the time being:
-> I refactored the function `fetch_latest_prompt` to use the `session_identifier` from the instance if that's not `None`, else use the default value. I think that's what your suggestion was.
-> For the future, my original idea (which I still think is good) is to have `None` as session_identifier all the way until the `Prompt` gets written, and the DB has a default value of PROMPT_DEFAULT_SESSION_IDENTIFIER. That is also used when fetching things. This avoids default arguments being passed multiple times and bugs as you discovered above.

prediction_market_agent/db/db_storage.py
gabrielfior: done

PR #239, opened 2024-06-11
Remove cap on bet amount for microchain agent

evangriffiths:
Just removing the bet cap meant the agent would still sometimes place very large bets (i.e. its full wallet!). I've added a new function, `GetKellyBet`. It looks like this is enough to get the agent to bet less rashly, but also more optimally.

Screenshot 2024-07-31 at 12 46 25


gabrielfior (approved):

prediction_market_agent/agents/microchain_agent/functions.py
```diff
@@ -160,9 +160,6 @@ def __init__(self, market_type: MarketType, outcome: str):
         )
         self.user_address = APIKeys().bet_from_address
 
-        # Prevent the agent from spending recklessly!
-        self.MAX_AMOUNT = 0.1 if market_type == MarketType.OMEN else 1.0
```
kongzii, line 164: YOLOOO

kongzii (approved):

evangriffiths:
Results from testing the agent without the cap:

- With the default trader-agent prompt, it still bets v large amounts (emptying the wallet in a few bets)
- With the 'just-born' config I tested, it was more conservative in its bet sizes (maybe due to 'survival' being mentioned in the prompt:
```
Act as an autonomous agent with a goal to learn, survive and evolve. AI agents learn by using learning functions and then updating their prompt using the combination of first getting their prompt, then modifying it and then updating it. After each Learn function called, the agent should update its prompt to include the new knowledge. The new knowledge needs to be about what was actually learned, not about just describing it. Updating the prompt after a learning session is an absolute necessity to make sure that the agent is able to use the new knowledge in the future. Agents should keep a track record of what they have learned in the past, to not repeat the same learning session again and again. A good prompt is clear, concise, and informative. It should contain all necessary information for the agent to perform well. Instead of saying 'I know about X,' the prompt should provide actual knowledge about X, including what it is and how it works. Reasoning is the process of thinking about something in a logical way to form a conclusion or judgment. The Reasoning function should be used often to help guide actions. Survival is the state or fact of continuing to live or exist. For an AI agent to survive, it must be able to continue to operate and perform its functions. In order to operate, it must keep a positive balance in its account. Every action or function that the agent does costs some amount of money. By acting as a trader agent in prediction markets and using other available functions to operate on them, an AI agent can create a profit to keep its balance positive.
```

I'll try changing the default trader-agent prompt and see if this helps.

evangriffiths:
> I'll try changing the default trader-agent prompt and see if this helps

Fixed by adding a new MarketFunction `GetKellyBet` instead.

PR #244, opened 2024-06-12
Gabriel/add graph api key

gabrielfior:


evangriffiths:
Also need to add the secret to the env section of python_ci.yaml https://github.com/gnosis/prediction-market-agent/blob/main/.github/workflows/python_ci.yaml#L9

evangriffiths (approved):

PR #254, opened 2024-06-17
Fix system prompt/bootstrap/function list lag in microchain app

evangriffiths:


kongzii (approved):
Looks better indeed! Thanks!

PR #259, opened 2024-06-19
242 unify dbstorage classes one per table

gabrielfior:


prediction_market_agent/agents/microchain_agent/memory.py
```diff
@@ -63,35 +61,3 @@ def from_long_term_memory(
             ),
             datetime_=long_term_memory.datetime_,
         )
-
-
-class LongTermMemory:
-    def __init__(self, task_description: str, sqlalchemy_db_url: str | None = None):
-        self.task_description = task_description
-        self.storage = DBStorage(sqlalchemy_db_url=sqlalchemy_db_url)
```
gabrielfior, line 71: Moved to `LongTermMemoryTableHandler`

prediction_market_agent/db/long_term_memory_table_handler.py
```diff
@@ -0,0 +1,58 @@
...
+            col(LongTermMemories.task_description) == self.task_description
+        ]
+        if from_ is not None:
+            query_filters.append(col(LongTermMemories.datetime_) >= from_)
+        if to_ is not None:
+            query_filters.append(col(LongTermMemories.datetime_) <= to_)
+
+        return self.sql_handler.get_with_filter_and_order(
+            query_filters=query_filters,
+            order_by_column_name=LongTermMemories.datetime_.key,  # type: ignore
```
gabrielfior, line 56: In multiple instances,  mypy does not play nicely with SQLModel columns. It defaults to the underlying datatype, instead of understanding it's also a column. Hence the `type ignore`. Suggestions welcome!

prediction_market_agent/db/sql_handler.py
```diff
@@ -0,0 +1,56 @@
+from loguru import logger
+from prediction_market_agent_tooling.tools.utils import check_not_none
+from sqlalchemy import Table, ColumnElement, Column, BinaryExpression
+from sqlmodel import create_engine, SQLModel, Session, select, desc, col, asc
+
+# from prediction_market_agent.db.models import LongTermMemories, Prompt
+import typing as t
+from prediction_market_agent.utils import DBKeys
+
+TypeTable = t.TypeVar("TypeTable", bound=SQLModel)
```
gabrielfior, line 10: This is needed otherwise using `Prompt` or `LongTermMemory` as return types of a downstream method doesn't match `SQLModel` and mypy complains.

prediction_market_agent/agents/microchain_agent/memory.py
```diff
@@ -59,126 +57,3 @@ def from_long_term_memory(
...
-        to: datetime | None = None,
-    ) -> Sequence[LongTermMemories]:
-        return self.storage.load_long_term_memories(
-            task_description=self.task_description,
-            from_=from_,
-            to=to,
-        )
-
-
-class ChatHistory(BaseModel):
```
evangriffiths, line 100: Looks like these new classes have been deleted by mistake?

prediction_market_agent/db/long_term_memory_table_handler.py
evangriffiths: Hmm that's annoying. Not sure about a fix. Couldn't find anything from a quick search - seems like a limitation of mypy as it's a static type checker, and this is a dynamic attribute.

gabrielfior: Yep, can't think of a good solution either

PR #265, opened 2024-06-21
Remove bootstrap

kongzii:


evangriffiths:
Great. Does it appear to behave the same before/after?

evangriffiths (approved):

Issue #271, opened 2024-06-25
Add `CallAPI` function to general agent


kongzii:
Add a universal `CallAPI` function to a general agent, which will just use the `requests` library to make arbitrary calls. To any endpoint, with any parameters, etc.

vishu1089:
on it!

vishu1089:
Yes, I would love to!

kongzii:
@vishu1089 Cool!

As an example, you can [take this action](https://github.com/gnosis/prediction-market-agent/blob/5304d2cb043072407627e5e5323436c45b74d052/prediction_market_agent/agents/microchain_agent/memory_functions.py#L13).

You can copy-paste and

1. Change the description accordingly
2. Change the example_args and __call__ function to accept `url` and `params` as input arguments
3. It should do a API call with requests library and return it's response (as string).

How exactly it will look like is open question, depends on what will work with LLM and the microchain framework.

For testing, you can[ add the new function here](https://github.com/gnosis/prediction-market-agent/blob/5304d2cb043072407627e5e5323436c45b74d052/prediction_market_agent/agents/microchain_agent/microchain_agent.py#L35) and then try to prompt the agent to use [it in the streamlit demo here.](https://github.com/gnosis/prediction-market-agent/blob/5304d2cb043072407627e5e5323436c45b74d052/prediction_market_agent/agents/microchain_agent/app.py).

Ping us if we can help along the way 🙏

vishu1089:
Thanks, I'll check it out.

vishu1089:
Got it, thanks!

vishu1089:
Need some context, lemme know when can we connect?

kongzii:
Closing as will be solved by https://github.com/gnosis/prediction-market-agent/pull/278

PR #278, opened 2024-06-26
#271 issue: Add call API function to general agent

vishu1089:
PR for #271

### DIFF
--- a/prediction_market_agent/agents/microchain_agent/call_api.py
+++ b/prediction_market_agent/agents/microchain_agent/call_api.py
@@ -23,7 +23,6 @@ class CallAPI:
         """
         Returns a list of strings representing the example arguments for the API call.
 
-        :rtype: list[str]
         """
         return ["method", "url", "params", "data", "headers"]
 
--- a/prediction_market_agent/agents/microchain_agent/call_api.py
+++ b/prediction_market_agent/agents/microchain_agent/call_api.py
@@ -1,0 +1,50 @@
+import requests
+
+class CallAPI:
+    def __init__(self) -> None:
+        """
+        Initializes the class instance.
+        """
+        super().__init__()
+
+    @property
+    def description(self) -> str:
+        """
+        A property method that returns a description of the function, which can be used to make arbitrary API calls to any endpoint with any parameters. It supports different HTTP methods (GET, POST, etc.) and can include headers, query parameters, and request bodies.
+        """
+        return (
+            "We can use this function to make arbitrary API calls to any endpoint with any parameters. "
+            "It supports different HTTP methods (GET, POST, etc.) and can include headers, "
+            "query parameters, and request bodies."
+        )
+
+    @property
+    def example_args(self) -> list[str]:
+        """
+        Returns a list of strings representing the example arguments for the API call.
+
+        :return: A list of strings representing the example arguments for the API call.
+        :rtype: list[str]
+        """
+        return ["method", "url", "params", "data", "headers"]
+
+    def __call__(self, method: str, url: str, params: dict = None, data: dict = None, headers: dict = None) -> str:
+        """
+        Sends an HTTP request using the specified method to the given URL with optional parameters, data, and headers.
+
+        Args:
+            method (str): The HTTP method to use for the request.
+            url (str): The URL to send the request to.
+            params (dict, optional): Query parameters to include in the request. Defaults to None.
+            data (dict, optional): Request body to include in the request. Defaults to None.
+            headers (dict, optional): Headers to include in the request. Defaults to None.
+
+        Returns:
+            str: The response string if the request was successful, or the error message if an exception occurred.
+        """
+        try:
+            response = requests.request(method, url, params=params, data=data, headers=headers)
+            response.raise_for_status()
+            return response.text
+        except requests.exceptions.RequestException as e:
+            return str(e)
--- a/prediction_market_agent/agents/microchain_agent/call_api.py
+++ b/prediction_market_agent/agents/microchain_agent/call_api.py
@@ -1,12 +1,7 @@
 import requests
-
-class CallAPI:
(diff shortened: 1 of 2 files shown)


prediction_market_agent/agents/microchain_agent/call_api.py
```diff
@@ -0,0 +1,50 @@
+import requests
+
+class CallAPI:
```
kongzii, line 3: CallAPI needs to inherit from Function, [for example as here](https://github.com/gnosis/prediction-market-agent/blob/5304d2cb043072407627e5e5323436c45b74d052/prediction_market_agent/agents/microchain_agent/memory_functions.py#L13).

prediction_market_agent/agents/microchain_agent/call_api.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self) -> None:
+        """
+        Initializes the class instance.
+        """
+        super().__init__()
```
kongzii, line 8: This whole `__init__` method can be removed, as nothing goes on there.

prediction_market_agent/agents/microchain_agent/call_api.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @property
+    def description(self) -> str:
+        """
+        A property method that returns a description of the function, which can be used to make arbitrary API calls to any endpoint with any parameters. It supports different HTTP methods (GET, POST, etc.) and can include headers, query parameters, and request bodies.
+        """
+        return (
+            "We can use this function to make arbitrary API calls to any endpoint with any parameters. "
```
kongzii, line 16:
```suggestion
            "Use this function to make arbitrary API calls to any endpoint with any parameters. "
```

prediction_market_agent/agents/microchain_agent/call_api.py
```diff
@@ -0,0 +1,50 @@
...
+            "It supports different HTTP methods (GET, POST, etc.) and can include headers, "
+            "query parameters, and request bodies."
+        )
+
+    @property
+    def example_args(self) -> list[str]:
+        """
+        Returns a list of strings representing the example arguments for the API call.
+
+        :return: A list of strings representing the example arguments for the API call.
```
kongzii, line 26:
```suggestion
```

No need to duplicate comments

prediction_market_agent/agents/microchain_agent/call_api.py
```diff
@@ -0,0 +1,50 @@
...
+        :rtype: list[str]
```
kongzii, line 27: We use python type annotations, so no need for this
```suggestion
```

prediction_market_agent/agents/microchain_agent/call_api.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        return ["method", "url", "params", "data", "headers"]
+
+    def __call__(self, method: str, url: str, params: dict = None, data: dict = None, headers: dict = None) -> str:
```
kongzii, line 31:
```suggestion
    def __call__(self, method: str, url: str, params: dict[str, Any] | None = None, data: dict[str, Any] | None = None, headers: dict[str, Any] | None = None) -> str:
```

prediction_market_agent/agents/microchain_agent/call_api.py
```diff
@@ -0,0 +1,50 @@
...
+
+        Returns:
+            str: The response string if the request was successful, or the error message if an exception occurred.
+        """
+        try:
+            response = requests.request(method, url, params=params, data=data, headers=headers)
+            response.raise_for_status()
+            return response.text
+        except requests.exceptions.RequestException as e:
+            return {"success": False, "error": str(e)}
```
kongzii, line 50: Could you remove the try-except block please, other tools will throw an exception as well, so to not fail silently.

vishu1089:
Super helpful!


Will take it as soon as possible!

prediction_market_agent/agents/microchain_agent/call_api.py
kongzii: Oh, for this to work, you will also need to add `from typing import Any` at the top of the file.

prediction_market_agent/agents/microchain_agent/call_api.py
vishu1089: okay

vishu1089:
review changes done, please check

vishu1089:
updated the branch with main

kongzii:
@vishu1089 Thanks! Mostly perfect, but I pushed a few small tweaks (I didn't know what's wrong, and at the time I figured it out, it was basically done 😄 )

Tweaks include:

1. Run the code with black, autoflake and isort formatters ([see our pipeline](https://github.com/gnosis/prediction-market-agent/blob/918cf8ddfb870568b65d9bf4b5a719ed2521315b/.github/workflows/python_ci.yaml#L62))
2. Add the new Function tool to the agent, so he can use it
3. Fix the tool, because it didn't work with Microchain framework if input parameters were dictionaries, it need to be a string and then converted to the dictionary inside of the tool. (please, t[est out new functions using the app I suggested in the issue](https://github.com/gnosis/prediction-market-agent/issues/271#issuecomment-2188579007))
4. For some reason, microchain framework doesn't like typing using `|`, so I switched it to the old school `Optional[...]` instead.

I'll wait for the CI to pass and merge if all is okay.

Thanks for the help!

Screenshot by Dropbox Capture

kongzii (approved):

kongzii (approved):

kongzii:
Some tests failed, but it's unrelated to this PR. Seems like secrets aren't available in PRs opened from forks (which makes sense). We need to look into how to fix that, so merging this now (cc @evangriffiths @gabrielfior ):

Screenshot by Dropbox Capture

PR #283, opened 2024-06-26
Fix image generation in replicator

kongzii:


prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
```diff
@@ -133,11 +133,10 @@ def omen_replicate_from_tx(
             f"Created `https://aiomen.eth.limo/#/{market_address}` for `{market.question}` in category {category} out of {market.url}."
         )
 
-        omen_agent_market = OmenAgentMarket.from_data_model(
-            OmenSubgraphHandler().get_omen_market_by_market_id(market_address)
-        )
-
-        if generate_and_set_image_for_market(omen_agent_market, api_keys) is not None:
+        if (
```
kongzii, line 136: Because of the graph latency, sometimes it happens that the created market is not yet indexed in the graph and then replication fails, because `OmenSubgraphHandler().get_omen_market_by_market_id(market_address)` will error out.

So instead, provide just the address and question directly.

prediction_market_agent/agents/replicate_to_omen_agent/image_gen.py
```diff
@@ -16,21 +16,26 @@
 
 
 def generate_and_set_image_for_market(
-    market: OmenAgentMarket, api_keys: APIKeys
+    market_address: ChecksumAddress, market_question: str, api_keys: APIKeys
 ) -> IPFSCIDVersion0 | None:
+    market_contract = OmenFixedProductMarketMakerContract(address=market_address)
+    # Test that the market actually exists.
+    # TODO: How can we get the question from the market contract (without using the graph)? (to check that `market_question` is correct)
+    market_contract.totalSupply()
```
evangriffiths, line 24: Delete this code?

kongzii: It will fail if `market_address` doesn't exist, is there a better way to test it than calling a random method?

evangriffiths: oh i seeeee, makes sense. I don't know if there's a better way to test 😢. But you could do a try-except block, and if it fails, throw a more descriptive exception, like `"Unable to verify contract exists for market address {addr}"`

evangriffiths (approved):

prediction_market_agent/agents/replicate_to_omen_agent/image_gen.py
kongzii: True! Added.

PR #285, opened 2024-06-26
Fix delployed microchain agent

evangriffiths:


prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -44,9 +47,12 @@ def run(
         )
         agent.run(self.n_iterations)
         long_term_memory.save_history(agent.history)
-        prompt_handler.save_prompt(agent.system_prompt)
+        editable_prompt = get_editable_prompt_from_agent(agent)
```
evangriffiths, line 50: Otherwise fails with `The system prompt already contains the non-updatable part.` inside `build_full_system_prompt` when loading a `historical_prompt`

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -107,7 +106,9 @@ def main(
 ) -> None:
     # This description below serves to unique identify agent entries on the LTM, and should be
     # unique across instances (i.e. markets).
-    unique_task_description = AgentIdentifier.microchain_task_from_market(market_type)
+    unique_task_description = (
```
evangriffiths, line 109: So that running locally doesn't add memories under the same description as deployed agent

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -44,7 +47,9 @@ def run(
...
+        if prompt_handler:
```
kongzii, line 51: Seems like this will always be True

kongzii (approved):

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -138,7 +139,3 @@ def main(
 
 def get_editable_prompt_from_agent(agent: Agent) -> str:
     return extract_updatable_system_prompt(str(agent.system_prompt))
-
-
-if __name__ == "__main__":
```
kongzii, line 143: Was this deleted intentionally?

evangriffiths: Oops, it was originally but I will re-instate it

prediction_market_agent/agents/microchain_agent/deploy.py
evangriffiths: 😱

PR #295, opened 2024-07-01
Add remaining days to the TT agent

kongzii:


prediction_market_agent/agents/utils.py
```diff
@@ -108,3 +110,26 @@ def memories_to_learnings(memories: list[DatedChatMessage], model: str) -> str:
...
+
+def get_event_date_from_question(question: str) -> datetime | None:
+    llm = ChatOpenAI(
+        model="gpt-4-turbo",
+        temperature=0.0,
+        api_key=APIKeys().openai_api_key.get_secret_value(),
+    )
+    event_date_str = str(
+        llm.invoke(
+            f"Extract the event date in the format `%m-%d-%Y` from the following question, don't write anything else, only the event date in the given format: `{question}`"
```
kongzii, line 123: We need to extract this from the question itself, instead of looking at market's closing date, because on Omen, recommendation is to close market 7 days after the outcome is known.

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
```diff
@@ -100,6 +100,8 @@
  
 - You should determine the probability of the SCENARIO SCENARIO_TO_ASSESS being true, considering the probabilities of the other related SCENARIOs.
 
+- Keep in mind that the scenario needs to happen in {n_remaining_days} days and the market is already open for {n_market_open_days} days. If it didn't happen yet, consider how many days are left relative to the number of days the market is open.
```
gabrielfior, line 103: I would replace `happen` by `finish` or `be resolved`, to make it clear it should be resolved by date X.

prediction_market_agent/agents/utils.py
gabrielfior: Suggestion - use Pydantic output parser for safer parsing (https://python.langchain.com/v0.1/docs/modules/model_io/output_parsers/types/pydantic/).

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
kongzii: `finish` doesn't sound right to me, e.g. `Will X be arrested by end of July?` --> `It needs to finish in X days`.

`be resolved` can mean both `Yes` and `No` as a resolution; in that case, the probability is 100% because it will be somehow resolved for sure.

Could you rewrite the whole sentence so I can see how you mean it?

gabrielfior: @evangriffiths can you help us formulate this better?

evangriffiths: I think it's okay, but I see what you mean @gabrielfior. Maybe something like `Your answer should reflect the time sensitivity of the scenario - the market closes in `n_remaining_days`.

But also

> consider how many days are left relative to the number of days the market is open.

It's not clear to me why the answer to a question depends on `n_market_open_days`. e.g. `Will X be arrested by end of July?`. If you are answering today (2nd July), the answer would be the same if the market was opened yesterday vs 1 month ago.

kongzii: > Maybe something like Your answer should reflect the time sensitivity of the scenario - the market closes in n_remaining_days`.

Sounds good, I'll update it.

> It's not clear to me why the answer to a question depends on n_market_open_days. e.g. Will X be arrested by end of July?. If you are answering today (2nd July), the answer would be the same if the market was opened yesterday vs 1 month ago.

The idea is that it should not :o However, my assumption about when the market is opened was wrong. 

Let's say I ask you `Will X be arrested by end of July?`, I think anything between 0 and 100% is valid guess.

What if I ask you `Will X be arrested by end of July? They are looking for her for past 7 years`, I think valid guess is anything less than 1% 😄 .

The idea here is similar, if 28 days are remaining (`answering today (2nd July)`) until the market is closed as in your example, there are good chances she will get arrested.

If only 1 day is remaining (answering at the end of the July) and they still didn't arrest her, what are the chances it will happen?

---

However the logic in this PR isn't great, because in this case, markets was opened 2 days ago and closes in 1 day.  It would be more interesting if market was opened 7 years ago. So I created this issue to improve the logic from this PR: https://github.com/gnosis/prediction-market-agent/issues/296

evangriffiths: Sorry, I still don't understand why 'proportion of duration of market elapsed' should affect p_yes estimate.

> However, my assumption about when the market is opened was wrong.

What do you mean by this?

> They are looking for her for past 7 years

This is something that should be picked up by the agent in the research report, so it shouldn't matter when the market is created.

kongzii: Updated the prompt.

> Sorry, I still don't understand why 'proportion of duration of market elapsed' should affect p_yes estimate.

> This is something that should be picked up by the agent in the research report, so it shouldn't matter when the market is created.

It should, but it's not. `market created` is an approximation. And https://github.com/gnosis/prediction-market-agent/issues/296 is to make it better.

> What do you mean by this?

I thought that market like https://presagio.pages.dev/markets?id=0xec2d18818ae6311e113ae115027c1b12c3f4e3e5 would be created a long time ago, and so `market created` would be a good approximation for this. Unfortunatelly, it was created just a few days ago, so not great 🤷

evangriffiths: Ohh okay, so `market created` is meant to be a proxy for 'when to start the exponential curve in your probability model', or whatever?

Well I still think the `and the market was already open for {n_market_open_days} days.` bit should be removed, and done another way. But I don't have a good suggestion for now, so accepting :)

evangriffiths (approved):

prediction_market_agent/agents/think_thoroughly_agent/prompts.py
kongzii: > Well I still think the and the market was already open for {n_market_open_days} days. bit should be removed, and done another way. But I don't have a good suggestion for now

Same here! 😄

PR #305, opened 2024-07-03
Allow multiple general agents

kongzii:
Screenshot by Dropbox Capture


prediction_market_agent/agents/utils.py
```diff
@@ -18,16 +17,22 @@
 class AgentIdentifier(str, Enum):
     THINK_THOROUGHLY = "think-thoroughly-agent"
     MICROCHAIN_AGENT_OMEN = "microchain-agent-deployment-omen"
+    MICROCHAIN_AGENT_OMEN_TEST = "microchain-agent-deployment-omen_test"
+    MICROCHAIN_AGENT_OMEN_LEARNING_0 = "general-agent-0"
```
kongzii, line 21: Named them just `general-agent-X`, because it was too long for the select box 😄

scripts/deployed_general_agent_viewer.py
```diff
@@ -19,33 +18,41 @@
...
 from prediction_market_agent.agents.utils import AgentIdentifier
 from prediction_market_agent.db.long_term_memory_table_handler import (
     LongTermMemoryTableHandler,
 )
-from prediction_market_agent.tools.streamlit_utils import check_required_api_keys
 from prediction_market_agent.utils import APIKeys
 
 
-class DeployedGeneralAgentKeys(APIKeys):
-    START_TIME: datetime
```
kongzii, line 38: Deleted starting time, as we can get it easily from the chat history object below. So we don't need to track this in the env for all the general agents.

scripts/deployed_general_agent_viewer.py
```diff
@@ -57,10 +68,22 @@ def starting_balance(self) -> float:
 )
 st.title("Deployed Trader Agent Viewer")
 
-check_required_api_keys(["BET_FROM_PRIVATE_KEY"])
```
kongzii, line 60: `check_required_api_keys` not needed anymore, because now the requried env vars aren't optional in  DeployedGeneralAgentSettings

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -51,6 +51,19 @@ def run(
         prompt_handler.save_prompt(editable_prompt)
 
 
-class DeployableMicrochainModifiableSystemPromptAgent(DeployableMicrochainAgent):
+class DeployableMicrochainModifiableSystemPromptAgent0(DeployableMicrochainAgent):
     system_prompt_choice: SystemPromptChoice = SystemPromptChoice.JUST_BORN
     load_historical_prompt: bool = True
+    task_description = AgentIdentifier.MICROCHAIN_AGENT_OMEN_LEARNING_0
```
evangriffiths, line 57: This could be set via env var, and then you wouldn't need to duplicate this python code, and all the duplication would just be in the terraform modules

kongzii: I know but I actually liked that we have this explicit list of available agents: https://github.com/gnosis/prediction-market-agent/blob/8898be69854172abae0552ed2c5d3d037bbdeb53/prediction_market_agent/run_agent.py#L48

But I agree there is a ton of duplication, what do you think about it now? I did a small refactor.

prediction_market_agent/agents/utils.py
evangriffiths: ahh fair enough 😄

scripts/deployed_general_agent_viewer.py
evangriffiths: nice, that's better!

prediction_market_agent/agents/microchain_agent/deploy.py
evangriffiths: I'm not sure. My instinct is that multiple copies of the same agent should be specified only in the terraform code. But by that logic I should have made `load_historical_prompt` settable by env var too, and not have `DeployableMicrochainModifiableSystemPromptAgent` either. So happy to go with this for now :)

evangriffiths (approved):

kongzii:
Waiting for funding for these new agents by finance people, if they won't send the some xDai soonish I will invest in them by myself 😄

PR #307, opened 2024-07-04
263 thinkthoroughlyagents correlated markets feature has unexpected behaviour

gabrielfior:


prediction_market_agent/db/pinecone_handler.py
```diff
@@ -4,36 +4,51 @@
...
+    index: Index
+
+    def __init__(self, model: str = "text-embedding-3-large") -> None:
+        self.keys = APIKeys()
+        self.model = model
         self.embeddings = OpenAIEmbeddings(
-            api_key=keys.openai_api_key.get_secret_value()
+            api_key=SecretStrV1(self.keys.openai_api_key.get_secret_value()),
+            model=model,
         )
```
gabrielfior, line 41: Changed type to make mypy happy

prediction_market_agent/db/pinecone_handler.py
```diff
@@ -4,36 +4,51 @@
...
+INDEX_NAME = "omen-index-text-embeddings-3-large"
 T = t.TypeVar("T")
 
 
 class PineconeHandler:
-    def __init__(self) -> None:
-        keys = APIKeys()
-        self.pc = Pinecone(api_key=keys.pinecone_api_key.get_secret_value())
-        self.index = self.pc.Index(INDEX_NAME)
+    vectorstore: VectorStore
```
gabrielfior, line 31: Using base class to use an in-memory version during tests.

prediction_market_agent/db/pinecone_handler.py
```diff
@@ -77,36 +108,38 @@ def insert_all_omen_markets_if_not_exists(
...
-        self, limit: int, text: str, threshold: float = 0.7
+        self, limit: int, text: str, threshold: float = 0.25
     ) -> list[PineconeMetadata]:
         # Note that pagination is not implemented in the Pinecone client.
         # Hence we set a large limit and hope we get enough results that satisfy the threshold.
-        documents_and_scores = self.vectorstore.similarity_search_with_score(
-            query=text, k=int(limit * 5)
+        documents_and_scores = self.vectorstore.similarity_search_with_relevance_scores(
+            query=text, k=limit, score_threshold=threshold
         )
```
gabrielfior, line 139: `similarity_search_with_relevance_scores` is doing the trick for finding more relevant markets, together with the larger embedding

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -151,7 +152,7 @@ def _build_llm(self) -> BaseChatModel:
         #  https://github.com/gnosis/prediction-market-agent/issues/107)
         llm = ChatOpenAI(
             model=self.model,
-            api_key=keys.openai_api_key.get_secret_value(),
+            api_key=SecretStrV1(keys.openai_api_key.get_secret_value()),
```
evangriffiths, line 155: Is this used like this in all instances of `APIKeys().openai_api_key`?

If so, is `pydantic.types.SecretStr` not a valid type?
- If it is, you could just do `api_key=keys.openai_api_key`.
- If not, you could change the return type of `APIKeys.openai_api_key` to `pydantic.v1.types.SecretStr`

pyproject.toml
```diff
@@ -9,7 +9,8 @@ readme = "README.md"
 python = "~3.10.0"
 python-dotenv = "*"
 openai = "*"
-langchain = "^0.1.1"
+# ToDo - Update to 2.0
+langchain = "0.2.6"
```
evangriffiths, line 13: Avoid strict dependency requirements. Also, why the TODO?
```suggestion
langchain = "^0.2.6"
```

pyproject.toml
```diff
@@ -21,7 +22,7 @@ google-search-results = "*"
 pytest = "*"
 llama-index = "~0.9.0"
 duckduckgo-search = "*"
-crewai = "^0.30.11"
+crewai = "^0.35.3"
```
evangriffiths, line 25: Looks unrelated to this PR?

pyproject.toml
```diff
@@ -35,7 +36,7 @@ poetry = "^1.7.1"
 poetry-plugin-export = "^1.6.0"
 functions-framework = "^3.5.0"
 cron-validator = "^1.0.8"
-prediction-market-agent-tooling = { version = "^0.40.0", extras = ["langchain", "google"] }
+prediction-market-agent-tooling = { version = "^0.40.0", extras =  ["google"] }
```
evangriffiths, line 39: Hmm I think this could be a problem, if PMAT and PMA use different versions of langchain, as the langchain features from PMAT are used in this repo.

In this case, do we actually need `langchain-chroma` for the test? We already have the `chromadb` package. If we do need it, then I think it should be updated in PMAT first.

prediction_market_agent/db/pinecone_handler.py
```diff
@@ -77,36 +108,38 @@ def insert_all_omen_markets_if_not_exists(
...
                 logger.debug(f"Inserting {len(text_chunk)} into the vector database.")
-                self.insert_texts_if_not_exists(
-                    texts=text_chunk, metadatas=metadata_chunk
+                self.insert_texts(
+                    ids=ids_chunk, texts=text_chunk, metadatas=metadata_chunk
                 )
 
     def find_nearest_questions_with_threshold(
-        self, limit: int, text: str, threshold: float = 0.7
+        self, limit: int, text: str, threshold: float = 0.25
```
evangriffiths, line 133: magic numbers changing 😅

prediction_market_agent/db/pinecone_handler.py
```diff
@@ -4,36 +4,51 @@
...
+from pydantic.v1.types import SecretStr as SecretStrV1
 from tqdm import tqdm
 
 from prediction_market_agent.agents.think_thoroughly_agent.models import (
     PineconeMetadata,
 )
 from prediction_market_agent.utils import APIKeys
 
-INDEX_NAME = "omen-markets"
+# INDEX_NAME = "omen-markets"
```
evangriffiths, line 25: Remove commented out code

prediction_market_agent/db/pinecone_handler.py
```diff
@@ -4,36 +4,51 @@
...
+INDEX_NAME = "omen-index-text-embeddings-3-large"
 T = t.TypeVar("T")
 
 
 class PineconeHandler:
-    def __init__(self) -> None:
-        keys = APIKeys()
-        self.pc = Pinecone(api_key=keys.pinecone_api_key.get_secret_value())
-        self.index = self.pc.Index(INDEX_NAME)
+    vectorstore: VectorStore
```
evangriffiths, line 31: Could you give an explanation of the changes in the logic of this class that have improved the performance?

pyproject.toml
gabrielfior: Chroma is not the problem - problem is, Langchain >=0.2.6 is required for the embeddings to work.
The way I got around it is by bypassing the langchain requirement from PMAT - is updating langchain on PMAT the suggested approach?

prediction_market_agent/db/pinecone_handler.py
gabrielfior: Did some heuristic testing 😆

pyproject.toml
evangriffiths: > is updating langchain on PMAT the suggested approach?

In that case, yes

> Langchain >=0.2.6 is required for the embeddings to work.

What library needs this? `langchain_openai`?

prediction_market_agent/db/pinecone_handler.py
gabrielfior: Sure.
The problem was, I was chunking the texts (market titles) to insert (roughly 4,000 in buckets of 100). For every chunk, I was calling Pinecone and asking for the IDs there, and then doing a set(new_texts_to_add).difference(ids_already_in_db) -> https://github.com/gnosis/prediction-market-agent/pull/307/files#diff-708990d1fa3062d041926232998f54cd6a5116d1b02f18bb6a14aa520cae7894L58

Now I do the `set difference` in the beginning, before chunking. Once I removed markets already present in the DB, chunking can be done safely (also see function `deduplicate_markets` for the case where the same market_title occurs (we don't insert duplicates).

gabrielfior:
@evangriffiths do I recall correctly that you had a comment ref. fetching open or resolved markets when executing the Think Thoroughly agent? Can you repeat the rationale here please?

image

prediction_market_agent/db/pinecone_handler.py
evangriffiths: Thanks. What is the reason for chunking here?

evangriffiths:
> @evangriffiths do I recall correctly that you had a comment ref. fetching open or resolved markets when executing the Think Thoroughly agent? Can you repeat the rationale here please?

What's the context, sorry? I can't remember such a comment!

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
gabrielfior: implemented 2nd approach

prediction_market_agent/db/pinecone_handler.py
gabrielfior: Pinecone has very long latency when uploading too many texts, hence sending 100 every time is safer to not incur timeouts, retry, etc.

pyproject.toml
gabrielfior: good catch - also removed todo

pyproject.toml
gabrielfior: Yes, `langchain_openai` because of `OpenAIEmbeddings`

pyproject.toml
gabrielfior: Crewai v 0.30.11 complains about langchain dependency, while 0.35.3 works.

gabrielfior:
> > @evangriffiths do I recall correctly that you had a comment ref. fetching open or resolved markets when executing the Think Thoroughly agent? Can you repeat the rationale here please?
> 
> What's the context, sorry? I can't remember such a comment!

Solved in the standup - nothing to do here.

prediction_market_agent/db/pinecone_handler.py
```diff
@@ -20,20 +21,32 @@
...
         self.embeddings = OpenAIEmbeddings(
-            api_key=keys.openai_api_key.get_secret_value()
+            api_key=self.keys.openai_api_key,
+            model=model,
         )
+        self.build_pinecone()
+        self.build_vectorstore()
+
+    def build_pinecone(self) -> None:
+        self.pc = Pinecone(api_key=self.keys.pinecone_api_key.get_secret_value())
```
kongzii, line 44: Is there a reason not to initialize self.pc, self. index, self.vectorstore in `__init__`? If no, please move it there, if yes then they should be created as `self.pc = None` and then inited here, but that will create some unnecessary type-checking troubles.

prediction_market_agent/utils.py
```diff
@@ -33,7 +34,7 @@ def serp_api_key(self) -> SecretStr:
         )
 
     @property
-    def openai_api_key(self) -> SecretStr:
+    def openai_api_key(self) -> SecretStrV1:
```
kongzii, line 37: Please keep the `SecretStr` here and use https://github.com/gnosis/prediction-market-agent-tooling/blob/bac961f73de2b0dce8b0892a484834dadf3683db/prediction_market_agent_tooling/gtypes.py#L78 where necessary.

V1 is older, so make sense to use that one only with older libraries that aren't updated.

```suggestion
    def openai_api_key(self) -> SecretStr:
```

prediction_market_agent/utils.py
```diff
@@ -20,7 +21,7 @@ class DBKeys(BaseSettings):
 
 class APIKeys(APIKeysBase):
     SERP_API_KEY: t.Optional[SecretStr] = None
-    OPENAI_API_KEY: t.Optional[SecretStr] = None
+    OPENAI_API_KEY: t.Optional[SecretStrV1] = None
```
kongzii, line 24:
```suggestion
    OPENAI_API_KEY: t.Optional[SecretStr] = None
```

pyproject.toml
```diff
@@ -35,7 +34,7 @@ poetry = "^1.7.1"
 poetry-plugin-export = "^1.6.0"
 functions-framework = "^3.5.0"
 cron-validator = "^1.0.8"
-prediction-market-agent-tooling = { version = "^0.40.0", extras = ["langchain", "google"] }
+prediction-market-agent-tooling = { version = "^0.40.2", extras =  ["langchain", "google"] }
```
kongzii, line 37:
```suggestion
prediction-market-agent-tooling = { version = "^0.40.2", extras = ["langchain", "google"] }
```

pyproject.toml
```diff
@@ -61,6 +60,13 @@ langchain-pinecone = "^0.1.1"
 pinatapy-vourhey = "^0.2.0"
 plotly = "^5.22.0"
 
+
+
+
+
```
kongzii, line 66: Could you remove all these blank lines? 🙏

prediction_market_agent/db/pinecone_handler.py
gabrielfior: Reason is I didn't want to have PINECONE_API_KEY as a requirement for tests.
As it's currently implemented, I can simply mock `build_pinecone` with a `return None` function and it works. See https://github.com/gnosis/prediction-market-agent/pull/307/files#diff-a3fce0cf7d663446c86cdbb35e885419e6711a20305e8a2a1630b00948f1ac78R31-R34.

Agree with `self.pc = None` and etc.

kongzii (approved):

PR #310, opened 2024-07-09
Turn metaculus bot tournament script into a DeployableAgent

evangriffiths:
According to https://www.metaculus.com/aib/ Metaculus are releasing 5-10 questions/day over the course of a month. So this makes it too much of a pain to run manually. Making it deployable.


prediction_market_agent/agents/metaculus_agent/deploy.py
```diff
@@ -0,0 +1,81 @@
...
+            else:
+                answer = Answer(
+                    p_yes=Probability(0.5),
+                    decision=True,
+                    reasoning="Just a test.",
+                    confidence=0.5,
+                )
+
+            if answer is None:
+                logger.info("No answer was given. Skipping")
```
kongzii, line 76: Should it be `error` so we notice it in the logs? Or it's fine to just ignore them?

kongzii (approved):

prediction_market_agent/agents/metaculus_agent/deploy.py
evangriffiths: Good idea! Changed

PR #313, opened 2024-07-10
Filter out banned categories in replicator

kongzii:


prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
```diff
@@ -40,6 +46,8 @@
 # That is because at the closing time, the question will open on Realitio, and we don't want it to be resolved as unknown/invalid.
 # All replicated markets that close at N, needs to have closing time on Realition N + `EXTEND_CLOSING_TIME_DELTA`.
 EXTEND_CLOSING_TIME_DELTA = timedelta(days=6)
+# Banned categories are not allowed to be used for new markets on Omen.
+BANNED_CATEGORIES = ["Manifold Business Future"]
```
evangriffiths, line 50: I didn't understand from the thread - why is this category banned? Just because lots of questions in this category have titles that make an implicit assumption that 'manifold' if the subject?

kongzii: Yeah, and I thought that it's also not so interesting to copy markets about another PM platform.

evangriffiths (commented):
Nice that you've managed to add this feature. I don't mean to be too negative - but I don't think this is a good solution, as it seems like it overfits too much.

I would prefer one either:

1. The suggestion I made on the telegram about checking if the market description contains extra info that makes the title by itself not descriptive enough
2. Don't ban anything, and just accept that there might be bad markets. If it's actually not resolvable then we can just rely on the resolution mechanism to resolve it as invalid. Or is that a bad view? 😬

kongzii:
Fair enough, I'll try (1) and if not let's go with (2) 😄

evangriffiths:
> This won't happen, because replicator just replicates the resolution as well, so if it's resolved on Manifold, it's resolved on Omen.

Ohhhh crap - I didn't think of that! Well in that case I don't have another suggestion apart from (1)

PR #322, opened 2024-07-17
Fix 'iterations' metric in deployed agent viewer app

evangriffiths:
e.g. `general-agent-1` has run for 50 iterations (1 session), but currently its metric shows 101 iterations!!

Before:
Screenshot 2024-07-17 at 17 16 25

After:
Screenshot 2024-07-17 at 17 33 57

### DIFF
--- a/prediction_market_agent/agents/microchain_agent/memory.py
+++ b/prediction_market_agent/agents/microchain_agent/memory.py
@@ -79,13 +79,21 @@ class ChatHistory(BaseModel):
     def num_messages(self) -> int:
         return len(self.chat_messages)
 
+    @property
+    def is_empty(self) -> bool:
+        return self.num_messages == 0
+
+    @property
+    def iterations(self) -> int:
+        # Number of messages:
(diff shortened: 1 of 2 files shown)


prediction_market_agent/agents/microchain_agent/memory.py
```diff
@@ -79,14 +79,22 @@ def from_list_of_dicts(list_of_dicts: list[Dict[str, str]]) -> "ChatHistory":
...
+        #   -1 (for sys prompt)
+        #   divide by 2 (for user and assistant messages)
+        #   round down to nearest integer (in case the session ended with a failed function call)
+        return (self.num_messages - 1) // 2
+
 
 class DatedChatHistory(ChatHistory):
     chat_messages: Sequence[DatedChatMessage]
 
-    @property
```
evangriffiths, line 86: Moving from child class to base - makes more sense there 😃

kongzii (approved):

Issue #332, opened 2024-07-23
General agent Streamlit calls build_functions multiple times, it should call only once during start-up


gabrielfior:
-> File `prediction_market_agent/agents/microchain_agent/app.py` contains a dashboard
-> In the following code snippet, the function `get_function_bullet_point_list` is called

```
# Display its updated function list, system prompt
with intro_expander:
    st.markdown(
        "To start, click 'Run' to see the agent in action, or execute the "
        "agent with your own reasoning."
    )
    st.markdown("It is equipped with the following tools:")

    st.markdown(
        get_function_bullet_point_list(
            agent=st.session_state.agent, model=st.session_state.model
        )
        if agent_is_initialized()
        else "The agent is not initialized yet."
    )
```

-> This should in theory occur only once (namely, during agent's initialization), but is called only when the agent is already initialized. This has undesirable side effects, including the dynamic generation of classes for wxDAI contracts (which triggers OpenAI calls), among others.

In this issue, we should avoid re-building the agents list every time, instead doing it only once in the beginning.

PR #336, opened 2024-07-25
Do not rebuild function list every general-agent app run

evangriffiths:
Functions names are still retrieved dynamically from the agent's engine on each call (so e.g. if a new Function is registered, it shows up), but without rebuilding functions.

Fixes https://github.com/gnosis/prediction-market-agent/issues/332

Screenshot 2024-07-25 at 11 44 13

### DIFF
--- a/prediction_market_agent/agents/microchain_agent/app.py
+++ b/prediction_market_agent/agents/microchain_agent/app.py
@@ -31,7 +31,6 @@ from streamlit_extras.bottom_container import bottom
 from prediction_market_agent.agents.microchain_agent.memory import ChatHistory
 from prediction_market_agent.agents.microchain_agent.microchain_agent import (
     build_agent,
-    build_agent_functions,
     get_unformatted_system_prompt,
 )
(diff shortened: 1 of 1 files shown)


prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -148,17 +147,12 @@ def maybe_initialize_agent(model: str, unformatted_system_prompt: str) -> None:
...
-        agent=agent,
-        market_type=MARKET_TYPE,
-        keys=KEYS,
-        long_term_memory=st.session_state.long_term_memory,
-        allow_stop=ALLOW_STOP,
-        model=model,
-    ):
-        bullet_points += f"  - {function.__class__.__name__}\n"
+    for function in agent.engine.functions:
+        bullet_points += f"  - {function}\n"
```
evangriffiths, line 155: `agent.engine.functions` is a dict with function names (`str`s) as keys.

kongzii (approved):

PR #343, opened 2024-07-26
Make PredictProbabilityForQuestion use Nevermined-wrapped prediction-prophet endpoint

evangriffiths:
Not intending to merge as-is. Rather, a demo of what it would look like to use tools based on Nevermined-wrapped endpoints as apposed to local code (or the mech-client)

This is an intermediate step towards the 'marketplace of tools' idea. The longer term vision would be for the agent to search the market for 'prediction' tools, and pick and use a relevant one.

Screenshot 2024-07-26 at 19 03 30
It works!


prediction_market_agent/tools/endpoints/prediction_prophet/deployment.py
```diff
@@ -0,0 +1,87 @@
...
+    p_yes: float
+    cost_usd: float
+
+
+def usd_to_credits(cost_usd: float) -> int:
+    # 1 USD = 100 credits
+    return math.ceil(cost_usd * 100)
+
+
+def predict(question: str) -> Response:
```
evangriffiths, line 29: Ah we have our own endpoint deployments in https://github.com/gnosis/labs-api/, so this would be moved there.

prediction_market_agent/tools/endpoints/prediction_prophet/nevermined.py
```diff
@@ -0,0 +1,45 @@
...
+Headers: TypeAlias = Dict[str, str]
+
+NAME = "Prediction Prophet"
+DESCRIPTION = (
+    "Ask a yes/no question about a future event, and get a prediction of the "
+    "probability of a 'yes' outcome. Uses Tavily web search and LLMs to perform "
+    "research, and turn the research into a prediction."
+)
+
+# Service costs between 0 and 20 credits (i.e. between $0.01 and $0.20)
```
kongzii, line 19: How is the actual cost for some call determined?

kongzii (commented):
Well seems to be simple enough 😄 (and all the parts actually make sense)

evangriffiths:
Closing until we decide to pursue the MoT idea one way or the other.

PR #353, opened 2024-08-01
Fix 'empty ChatHistory' bugs

evangriffiths:
If you view a deployed agent with an empty chat history, you hit:

Screenshot 2024-08-01 at 12 23 32


kongzii (approved):

prediction_market_agent/agents/microchain_agent/memory.py
```diff
@@ -93,7 +93,10 @@ def iterations(self) -> int:
         #   -1 (for sys prompt)
         #   divide by 2 (for user and assistant messages)
         #   round down to nearest integer (in case the session ended with a failed function call)
-        return (self.num_messages - 1) // 2
+        if self.is_empty:
+            return 0
+        else:
+            return (self.num_messages - 1) // 2
```
evangriffiths, line 99: otherwise you get `Number of iterations: -1` if empty

PR #360, opened 2024-08-02
Tasks PoC

kongzii:


prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -130,6 +143,9 @@ def build_agent(
     market_type: MarketType,
     model: SupportedModel,
     unformatted_system_prompt: str,
+    inc_trading_functions: bool,
+    inc_learning_functions: bool,
+    inc_universal_functions: bool,
```
evangriffiths, line 148: Don't feel suuuper strongly about it because maybe this is temporary anyway, but pass in a `FunctionsConfig` class to wrap these booleans, e.g.

```
class FunctionsConfig(BaseModel):
    include_trading_functions: bool
    include_learning_functions: bool
    include_universal_functions: bool

    @staticmethod
    def from_system_prompt_choice(system_prompt_choice: SystemPromptChoice):
        include_trading_functions = False
        include_learning_functions = False
        include_universal_functions = False

        if system_prompt_choice == SystemPromptChoice.JUST_BORN:
            include_learning_functions = True
            include_trading_functions = True

        elif system_prompt_choice == SystemPromptChoice.TRADING_AGENT:
            include_trading_functions = True

        elif system_prompt_choice == SystemPromptChoice.TASK_AGENT:
            include_universal_functions = True
            include_trading_functions = True

        return FunctionsConfig(
            include_trading_functions=include_trading_functions,
            include_learning_functions=include_learning_functions,
            include_universal_functions=include_universal_functions,
        )
```

so that:

- when you want to add a new boolean, you can change it just in that class, and don't have to change the code everywhere `build_agent` is called.
- that extra logic doesn't really belong in the `SystemPromptChoice` definition, and with this it can be removed.

Also I prefer `include_...` but up to you

kongzii: I like that, implemented

evangriffiths (approved):

PR #362, opened 2024-08-02
Add comment to exaplain `n_iterations` arg to `ThinkThoroughlyBase.answer_binary_market`

evangriffiths:
Otherwise if `n_iterations` > 1, `scenarios_with_probs` will only contain the scenarios for the final iteration

### DIFF
--- a/prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
+++ b/prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
@@ -338,7 +338,6 @@ class ThinkThoroughlyBase(ABC):
                 executor=DEFAULT_PROCESSPOOL_EXECUTOR,
             )
 
-            scenarios_with_probs = []
             for scenario, prediction in sub_predictions:
                 if prediction is None:
                     logger.warning(f"Could not generate prediction for '{scenario}'.")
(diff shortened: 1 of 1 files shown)


prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -338,7 +338,6 @@ def answer_binary_market(
                 executor=DEFAULT_PROCESSPOOL_EXECUTOR,
             )
 
-            scenarios_with_probs = []
```
kongzii, line 341: That's a little undocumented feature, actually 😄 Sorry

The idea is that:

First iteration, let's say scenarios_with_probs: a=80%, b=70%
Second iteration, scenarios_with_probs: a=40%, b=50%
Third iteration, scenarios_with_probs: a=45%, b=55%

So, the LLM is supposed to iteratively adjust the probs.

Without this resetting, it will be:

First iteration, let's say scenarios_with_probs: a=80%, b=70%
Second iteration, scenarios_with_probs: a=80%, b=70%,a=40%, b=50%
Third iteration, scenarios_with_probs: a=80%, b=70%,a=40%, b=50%,a=45%, b=55%

evangriffiths: oooh doh. the scenarios are outside of the for loop. I've added a comment so it's more obvious to future-me :)

kongzii (approved):

PR #368, opened 2024-08-06
Suppress transformers import error in Dockerfile

evangriffiths:
```
import subprocess
process = subprocess.run(
    ["python", "-c", "import transformers"],
    capture_output=True,
    text=True,
)
stderr_output = process.stderr
print("Captured stderr:", stderr_output)
```
Gives
```
Captured stderr: None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
```
because we don't have torch/tensorflow./flax as a dependency of this project.

We get this when we init a `ReplicateLlama31ChatGenerator` instance, and this shows up as an error in GCP logs
Screenshot 2024-08-06 at 10 35 36

Fix this by suppressing the warning in the Dockerfile by setting an env var.


Dockerfile
```diff
@@ -29,5 +29,9 @@ COPY scripts ./scripts
 
 ENV PYTHONPATH=/app
 ENV PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
+# Disable warning on imorting `transformers`, as we don't use torch in this
```
kongzii, line 32:
```suggestion
# Disable warning on importing `transformers`, as we don't use torch in this
```

kongzii (approved):

PR #383, opened 2024-08-07
Set minimum scraping requirement for prediction prophet research

evangriffiths:
Switch to latest prediction-prophet version, where failed web scrapes are not logged as errors. But to ensure bad scraping performance is not missed, add a minimum threshold for number of scraped websites


prediction_market_agent/tools/prediction_prophet/research.py
```diff
@@ -0,0 +1,39 @@
...
+        goal=goal,
+        model=model,
+        use_summaries=False,
+        initial_subqueries_limit=initial_subqueries_limit,
+        subqueries_limit=subqueries_limit,
+        max_results_per_search=max_results_per_search,
+        min_scraped_sites=min_scraped_sites,
+        openai_api_key=openai_api_key,
+        tavily_api_key=tavily_api_key,
+        logger=logger,
```
evangriffiths, line 38: use our own logger!

gabrielfior (approved):

prediction_market_agent/tools/prediction_prophet/research.py
kongzii: Note that our logger is used to show messages in Strealit apps: https://github.com/gnosis/prediction-market-agent/blob/f55a4648748602372630ba383c6e1708a5d11568/prediction_market_agent/tools/streamlit_utils.py#L49

So if this is changed, all the messages logged in the prophet_research will be shown as well, which I don't mind if they are or aren't, just an side-effect to have in mind.

kongzii (approved):

prediction_market_agent/tools/prediction_prophet/research.py
evangriffiths: Ah good point - I hadn't considered that! Had a look with `streamlit run scripts/agent_app.py` and I prefer it with the extra logging - and it made me see that we don't use our logger for the research calls in `DeployablePredictionProphet*Agent`s (because they call `research` indirectly, and that call is inside the predictionprophet library), which makes for a bit of an empty response in the app
Screenshot 2024-08-08 at 19 22 59

Made this issue to track https://github.com/gnosis/prediction-market-agent/issues/391 (can't do it in this PR now because it requires a change to predictionprophet).

PR #394, opened 2024-08-12
Retire gpt3.5 agents, replace with gpt4o

evangriffiths:
Through the API, these models are cheaper and perform better

### DIFF
--- a/prediction_market_agent/agents/known_outcome_agent/known_outcome_agent.py
+++ b/prediction_market_agent/agents/known_outcome_agent/known_outcome_agent.py
@@ -146,9 +146,7 @@ def summarize_if_required(content: str, model: str, question: str) -> str:
     """
     If the content is too long to fit in the model's context, summarize it.
     """
-    if model == "gpt-3.5-turbo-0125":  # 16k context length
-        max_length = 10000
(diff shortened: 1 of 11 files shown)


kongzii (approved):

prediction_market_agent/agents/known_outcome_agent/known_outcome_agent.py
```diff
@@ -146,9 +146,7 @@ def summarize_if_required(content: str, model: str, question: str) -> str:
     """
     If the content is too long to fit in the model's context, summarize it.
     """
-    if model == "gpt-3.5-turbo-0125":  # 16k context length
-        max_length = 10000
-    elif model == "gpt-4-1106-preview":  # 128k context length
+    if model == "gpt-4-1106-preview":  # 128k context length
```
gabrielfior, line 149: Why can't we use here also DEFAULT_MODEL from utils?

evangriffiths: If a new version of 4o is released that is cheaper+better, but has 256k context length, then we might change `DEFAULT_MODEL`, but we don't want to forget about updating this too.

PR #398, opened 2024-08-15
Langfuse integration for traders

kongzii:


prediction_market_agent/agents/microchain_agent/market_functions.py
```diff
@@ -123,19 +124,20 @@ def __init__(
     def description(self) -> str:
         return self._description
 
+    @observe(name="PredictProbabilityForQuestion")
```
kongzii, line 127: I created an issue in Microchain repo, to see if they are interested in having Langfuse support: https://github.com/galatolofederico/microchain/issues/16

If yes, I'll implement it, but in the meantime, track at least these predictions.

prediction_market_agent/agents/prophet_agent/deploy.py
```diff
@@ -14,22 +13,73 @@
...
     OlasAgent,
+    Prediction,
     PredictionProphetAgent,
+    Research,
 )
 
 from prediction_market_agent.utils import DEFAULT_OPENAI_MODEL
 
 
+class PredictionProphetAgentObserved(PredictionProphetAgent):
```
kongzii, line 32: Subclasses Prophet's agents just to `@observe` the methods. I thought it doesn't make sense to force `@observe` in Prophet's repository.

prediction_market_agent/agents/replicate_to_omen_agent/deploy.py
```diff
@@ -49,8 +51,14 @@ def run(self, market_type: MarketType = MarketType.MANIFOLD) -> None:
         if market_type != MarketType.OMEN:
             raise RuntimeError("Can replicate only into Omen.")
 
-        keys = APIKeys()
         settings = ReplicateSettings()
+        self.replicate(settings)
+
+    @observe()
+    def replicate(self, settings: ReplicateSettings) -> None:
+        self.langfuse_update_current_trace(tags=[REPLICATOR_TAG])
```
kongzii, line 59: Replicator isn't a trader agent, so we need to `@observe` and add tag manually.

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -383,8 +400,13 @@ def generate_prediction_for_one_outcome(
             list[tuple[str, AnswerWithScenario]] | None
         ) = None,
     ) -> AnswerWithScenario | None:
-        researcher = ThinkThoroughlyWithItsOwnResearch._get_researcher(model)
-        predictor = ThinkThoroughlyWithItsOwnResearch._get_predictor(model)
+        # Do not enable add_langfuse_callback, because it's not thread-safe.
```
kongzii, line 403: This is unfortunate, `generate_prediction_for_one_outcome` is executed in parallel, so we can't just decorate it with observe.  But we do track at least inputs and outputs.

pyproject.toml
```diff
@@ -34,12 +34,12 @@ poetry = "^1.7.1"
 poetry-plugin-export = "^1.6.0"
 functions-framework = "^3.5.0"
 cron-validator = "^1.0.8"
-prediction-market-agent-tooling = { version = "^0.47.1", extras = ["langchain", "google"] }
+prediction-market-agent-tooling = { path = "../prediction-market-agent-tooling", extras = ["langchain", "google"], develop = true }
```
kongzii, line 37: Needs to be updated

pyproject.toml
```diff
@@ -51,7 +51,8 @@ tweepy = "^4.14.0"
 langchain-pinecone = "^0.1.1"
 pinatapy-vourhey = "^0.2.0"
 plotly = "^5.22.0"
-prediction-prophet = { git = "https://github.com/agentcoinorg/predictionprophet.git", rev = "1ad9c99f519e3cc32570c36f5ed2e3de75b2b5f8" }
+# prediction-prophet = { git = "https://github.com/agentcoinorg/predictionprophet.git", rev = "1ad9c99f519e3cc32570c36f5ed2e3de75b2b5f8" }
+prediction-prophet = { path = "../evo.researcher", develop = true }
```
kongzii, line 55: Needs to be updated

prediction_market_agent/agents/known_outcome_agent/known_outcome_agent.py
```diff
@@ -178,7 +182,10 @@ def has_question_event_happened_in_the_past(
         date_str=date_str,
         question=question,
     )
-    answer = str(llm.invoke(prompt).content)
+    config: RunnableConfig = {
+        "callbacks": [langfuse_context.get_current_langchain_handler()]
+    }
+    answer = str(llm.invoke(prompt, config=config).content)
```
gabrielfior, line 188: It's not clear to me what `callbacks` mean, since they are being passed to the LLM definition (see line 177) and then being overwritten when `llm.invoke` is called (line 188).

prediction_market_agent/agents/prophet_agent/deploy.py
```diff
@@ -14,22 +13,73 @@
...
+
+class OlasAgentObserved(OlasAgent):
+    @observe()
+    def research(
+        self, market_question: str, add_langfuse_callback: bool = False
+    ) -> str:
+        return super().research(
+            market_question=market_question,
+            add_langfuse_callback=add_langfuse_callback,
+        )
```
gabrielfior, line 64: Having these extra subclasses just for Langfuse could be avoided (see https://github.com/gnosis/prediction-market-agent-tooling/pull/345/files#r1718609063). But also OK if done like this.

prediction_market_agent/agents/known_outcome_agent/known_outcome_agent.py
kongzii: Yep seems like they can be passed in both places and then they are merged:

Screenshot by Dropbox Capture

(this is from Langchain's code, where `callbacks` is from invoke's arguments, and `self.callbacks` is from the class)

I removed one of them, as we don't need both.

prediction_market_agent/agents/prophet_agent/deploy.py
kongzii: Yes, removed.

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -319,6 +332,9 @@ def answer_binary_market(
         hypothetical_scenarios = self.get_hypohetical_scenarios(question)
         conditional_scenarios = self.get_required_conditions(question)
 
+        unique_id = uuid4()
+        observe_unique_id(unique_id)
```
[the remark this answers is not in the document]
kongzii: I partially agree with both the concerns and the counterarguments. Just to clarify them a bit:

> not just isolated to some base class

The base code is isolated in the PMAT, if you create a new agent, then by default, you will get auto-observe on these, without any specific stuff in your agent:

Screenshot by Dropbox Capture

(so the base methods implemented on DeployableAgent class)

> the code for it is everywhere

Unfortunately, this is true as well. The basic observability from PMAT is great, but each agent has some specific flow (otherwise, it wouldn't be a separate agent, right?), and therefore, the observability code can not be abstracted away. For example, if we want to debug Prophet in the future, we need his research method to be observed.

This would be partially fixed [by this discussion I raised on Langfuse](https://github.com/orgs/langfuse/discussions/2949#discussioncomment-10358090). If they introduce an `auto-observe` feature, theoretically, we won't need to do this stuff manually in the agent's predicting logic. (doing this manually everywhere bothered me as well 😄 )

But even worse is that ThinkThoroughly needs multiprocessing to have a reasonable speed, and so the Langfuse integration is complicated because of it (complicated == sucks). The only solution there I can see is (1) do not use Langufse on him (not great), (2) do not use multiprocessing on him (also not great) 🤷 .

> if you don't want it for a particular agent, you don't need to add anything

This is true, but you will still get at least some basic observations logged, or you can disable it completely. 

> that's just how it is if you want observability

That is partially true; theoretically, we could agree that observing only basic methods from PMAT is enough and that we don't want more granular observations inside of the agent. But would that be really enough?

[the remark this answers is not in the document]
evangriffiths: Thanks for your response and making my concerns feel heard 😄 . I haven't got any suggestions for simplifying, so let's just go gettt itttttt.

evangriffiths (approved):

PR #407, opened 2024-08-20
Langfuse for General Agent!

kongzii:


pyproject.toml
```diff
@@ -40,7 +40,7 @@ autoflake = "^2.2.1"
 isort = "^5.13.2"
 markdownify = "^0.11.6"
 tavily-python = "^0.3.9"
-microchain-python = "^0.4.4"
+microchain-python = { path = "../microchain", develop = true }
```
kongzii, line 43: TODO: Needs change after Microchain is accepted & released

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -57,7 +60,9 @@
 ALLOW_STOP = False
 
 
+@observe()
 def run_agent(agent: Agent, iterations: int, model: SupportedModel) -> None:
+    langfuse_context.update_current_trace(tags=[GENERAL_AGENT_TAG, STREAMLIT_TAG])
```
evangriffiths, line 65: Does it make more sense to run this once at the start of the app, e.g. just after `ENABLE_LANGFUSE` is defined, you could do
```
if ENABLE_LANGFUSE:
    langfuse_context.update_current_trace(tags=[GENERAL_AGENT_TAG, STREAMLIT_TAG])
```

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -156,6 +157,7 @@ def build_agent(
             api_key=keys.openai_api_key.get_secret_value(),
             api_base=api_base,
             temperature=0.7,
+            enable_langfuse=enable_langfuse,
```
evangriffiths, line 160: put a comment on the microchain pr with an idea on how to remove the need to specify this flag here https://github.com/galatolofederico/microchain/pull/18/files#r1726981146

pyproject.toml
evangriffiths: in case you want to merge this before the microchain pr is merged, i guess you'll want to change this to your forked branch with a revision id 😺

prediction_market_agent/agents/microchain_agent/app.py
```diff
@@ -147,9 +153,19 @@ def maybe_initialize_agent(
             functions_config=FunctionsConfig.from_system_prompt_choice(
                 st.session_state.system_prompt_select
             ),
+            enable_langfuse=ENABLE_LANGFUSE,
+        )
+        # Use the un-observed version of these if Langfuse is enabled, because we don't want to have separate traces for these.
+        (
+            st.session_state.agent.reset()
+            if not ENABLE_LANGFUSE
```
evangriffiths, line 161: Do we not get the same behaviour in microchain as we have in pmat? If `ENABLE_LANGFUSE == False` then we don't do `agent.init_langfuse()`, so don't do `observe()(some_function)`, so this shouldn't be needed?

prediction_market_agent/agents/microchain_agent/app.py
kongzii: After the app is started, there is no observation yet. The first `@observe()` is on this `run_agent`.

However, it's not awesome that every click on "Run Agent" is a separate track in langfuse, but I at least added a session id, so they are grouped.

prediction_market_agent/agents/microchain_agent/microchain_agent.py
kongzii: It was a good idea, but I forgot that enable_langfuse property is required also because of this:

Screenshot by Dropbox Capture

So I just opened a PR fixing it 😄

prediction_market_agent/agents/microchain_agent/app.py
kongzii: In his library, I didn't put `@observe` on everything by default, but it's behind a flag: https://github.com/galatolofederico/microchain/blob/main/microchain/engine/agent.py#L33

kongzii: Maybe we don't need to track this streamlit application if it's bringing more complications than utility?

kongzii: Refactorer a bit, I introduced `total_iterations` state variable used as `resume` trigger:

```
agent.run(iterations=1, resume=st.session_state.total_iterations > 0)
```

so at the very first iteration, resume=False and `reset` + `build_initial_messages` are triggered inside of Microchain:

Screenshot by Dropbox Capture

evangriffiths (approved):

PR #409, opened 2024-08-21
Fix for Storage init

kongzii:


prediction_market_agent/agents/prophet_agent/deploy.py
```diff
@@ -86,32 +86,46 @@ def answer_binary_market(self, market: AgentMarket) -> Answer | None:
...
-        tavily_storage=TavilyStorage(agent_id="DeployablePredictionProphetGPT4oAgent"),
-        logger=logger,
-    )
+    agent: PredictionProphetAgent
+
+    def load(self) -> None:
+        super().load()
+        self.agent = PredictionProphetAgent(
+            model="gpt-4o-2024-08-06",
+            tavily_storage=TavilyStorage(agent_id=self.__class__.__name__),
```
kongzii, line 95: `TavilyStorage` can't be initialised on import-level, because then CoinFlip agent fails on SQL alchemy env variable missing, even when he doesn't need it at all

evangriffiths (approved):

PR #413, opened 2024-08-21
Add GoalManager, and make a DeployableMicrochainAgent with this feature

evangriffiths:
The `GoalManager` is a way of generating and evaluating goals for an agent. It can be used as a mechanism to learn from its experience, and add more structure to each session.

The motivation is that we in theory have a way for the agent to learn from its experience, through a combination of `RememberPastActions`+`UpdateMySystemPrompt`, but in practice for our deployed agents it's not clear how effective this is.  

The `Goal`s generated for an agent are lower-level than the description of the agent ("you are a PM trader agent, make money"), but higher-level than individual tools ("buy Yes tokens"). This makes them suited to proposing+evaluating on a per-session basis, and therefore (hopefully!) gives the agent some signal on how to adapt on a per-session basis.

An agent can use the `GoalManager` to:
- propose new goals (from scratch, or considering its goal history, including evaluations of whether they were completed)
- retry previous goals that were not completed
- evaluate whether the goal was completed
- save goals and their evaluations in a db

This feature is added to the microchain agent, but are default-off. A new deployable agent (`DeployableMicrochainWithGoalManagerAgent0`) is added for testing out this feature. You can see it in action below:

A goal is generated, and used as the user-prompt for the agent
Screenshot 2024-08-21 at 17 30 15

It seems to work quite nicely with the microchain agent without having to tweak any prompting. It reasons that it has completed the goal, and calls `Stop()` to terminate the program. The goal evaluation is appended to the chat history for the user to see.
Screenshot 2024-08-21 at 17 30 43


prediction_market_agent/agents/goal_manager.py
```diff
@@ -0,0 +1,300 @@
...
+        )
+        return [EvaluatedGoal.from_model(model) for model in evaluated_goal_models]
+
+    def generate_goal(self, latest_evaluated_goals: list[EvaluatedGoal]) -> Goal:
+        """
+        Generate a new goal based on the high-level description and the latest
+        evaluated goals.
+
+        TODO support generation of long-horizon goals with a specified
+        completion date, until which the goal's status is 'pending'.
```
evangriffiths, line 164: https://github.com/gnosis/prediction-market-agent/issues/415

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -104,3 +131,22 @@ class DeployableMicrochainModifiableSystemPromptAgent3(
...
+        high_level_description="You are a trader agent in prediction markets to maximise your profit.",
+        agent_capabilities=(
+            "You are able to:"
+            "\n- List all binary markets that can be traded."
+            "\n- List the current outcome probabilities for each open market."
+            "\n- Predict the outcome probability for a market."
+            "\n- Buy, sell and hold outcome tokens in a market."
+            "\n- Query your wallet balance, and the positions you hold in open markets."
+            "\n- Query the past bets you've made, and their outcomes."
+        ),
```
evangriffiths, line 150: I'm not super-keen on how the `high_level_description` and `agent_capabilities` are defined here. It feels like some duplication of the system prompt, and `engine.help` respectively. But I couldn't think of a better way of doing this
- `high_level_description` is different enough from the system prompt, and I didn't think it was a good solution to try and derive it from the system prompt via an llm call
- `agent_capabilities` could be derived from `agent.engine.functions`, but there would be some messiness if I tried to define this after `build_agent_functions` is called.

pyproject.toml
```diff
@@ -40,7 +40,7 @@ autoflake = "^2.2.1"
 isort = "^5.13.2"
 markdownify = "^0.11.6"
 tavily-python = "^0.3.9"
-microchain-python = "^0.4.4"
+microchain-python = { git = "https://github.com/galatolofederico/microchain.git", rev = "56db91de72b6466080c26434631808cc20af670c" }
```
evangriffiths, line 43: Using latest microchain revision until [this feature](https://github.com/galatolofederico/microchain/pull/17) makes its way into the release

prediction_market_agent/agents/microchain_agent/deploy.py
gabrielfior: This is good enough, but maybe you can use (as you mentioned) `agent.engine.functions` (or `agent.help`) to get a long list of functions, and have an LLM generate a summary of those and output a bullet-point list?
At least you stay up-to-date with the most recent functions.

prediction_market_agent/agents/goal_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        return (
+            f"# Goal:\n"
+            f"{self.goal}\n\n"
+            f"## Motivation:\n{self.motivation}\n\n"
+            f"## Completion Criteria:\n{self.completion_criteria}"
+        )
+
+
+class GoalEvaluation(BaseModel):
+    reasoning: str = Field(
```
evangriffiths, line 69: I saw on a random twitter comment that you should put the reasoning before the evaluation (`is_complete` in this case) in the prompt, so that the llm outputs the reasoning, then puts a score  (or here just a boolean) that's consistent with it, and not the other way around: https://x.com/dzhng/status/1823428375962407231

Not sure how true that is, but giving it a go here 😄 

(field ordering here is same same as in the formatting instructions of `EVALUATE_GOAL_PROGRESS_PROMPT_TEMPLATE`)

prediction_market_agent/agents/goal_manager.py
```diff
@@ -0,0 +1,295 @@
...
+            return True
+
+        if len(latest_evaluated_goals) < self.retry_limit + 1:
+            return False
+
+        latest_goal = latest_evaluated_goals[0].to_goal()
+        return all(
+            g.to_goal() == latest_goal
+            for g in latest_evaluated_goals[: self.retry_limit + 1]
+        )
```
gabrielfior, line 206: Could we use tenacity without decorators (see [link](https://github.com/jd/tenacity/issues/155)) here?

prediction_market_agent/agents/goal_manager.py
```diff
@@ -0,0 +1,295 @@
...
+            evaluated_goal = latest_evaluated_goals[0]
+            if evaluated_goal.is_complete:
+                # Generate a new goal
+                return self.generate_goal(latest_evaluated_goals)
+            else:
+                # Try again, unless we've reached the retry limit
+                if self.have_reached_retry_limit(latest_evaluated_goals):
+                    return self.generate_goal(latest_evaluated_goals)
+                else:
+                    return evaluated_goal.to_goal()
```
gabrielfior, line 229: Sorry, I'm a bit confused - how are you retrying? What constitutes an attempt - one iteration of the agent?
I don't love all these nested if-else, so wondering if we can simplify this.

prediction_market_agent/agents/goal_manager.py
kongzii: Yep, I witnessed this can help a lot

prediction_market_agent/agents/goal_manager.py
kongzii: What's wrong with the decorator?

prediction_market_agent/db/evaluated_goal_table_handler.py
```diff
@@ -0,0 +1,43 @@
...
+        sqlalchemy_db_url: str | None = None,
+    ):
+        self.agent_id = agent_id
+        self.sql_handler = SQLHandler(
+            model=EvaluatedGoalModel,
+            sqlalchemy_db_url=sqlalchemy_db_url,
+        )
+
+    def save_evaluated_goal(self, model: EvaluatedGoalModel) -> None:
+        self.sql_handler.save_multiple([model])
```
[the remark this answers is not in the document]
evangriffiths: Good idea! The current use case is to only generate one goal per session, so this wouldn't be used atm. But will keep in mind for the future!

prediction_market_agent/agents/goal_manager.py
evangriffiths: I'm not sure how tenacity.retry would fit into this. The goal manager here is fetching previous goals from the db, and checking if the latest goal is duplicated `self.retry_limit` times. i.e. there's no try-except loop to replace.

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -57,13 +70,27 @@ def run(
...
         initial_formatted_system_prompt = agent.system_prompt
 
         agent.run(self.n_iterations)
 
+        if self.goal_manager:
+            goal_evaluation = self.goal_manager.evaluate_goal_progress(
+                goal=goal,
+                chat_history=ChatHistory.from_list_of_dicts(agent.history),
+            )
+            self.goal_manager.save_evaluated_goal(goal=goal, evaluation=goal_evaluation)
```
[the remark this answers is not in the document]
evangriffiths: Do you prefer this way? It's meant an ugly `check_not_none(goal)` here to compensate.

prediction_market_agent/agents/goal_manager.py
gabrielfior: Understood that retry cannot be bundled with tenacity. Closing this.

prediction_market_agent/agents/goal_manager.py
gabrielfior: Clarified during stand-up -> retrying is checked/done after general agent has completed N iterations, after which goal is saved in the DB.
So flow is
-> create goal (contains logic for `retry` and `fetching same goal if it wasn't yet completed`)
-> do N iterations
-> store goal
-> create goal (same as above)
-> do N iterations
...
@evangriffiths can you confirm my understanding here?
If correct, please close this thread.

evangriffiths: Yes that's right! Sorry for leaving you hanging 🙏

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -28,6 +34,7 @@ class DeployableMicrochainAgent(DeployableAgent):
     load_historical_prompt: bool = False
     system_prompt_choice: SystemPromptChoice = SystemPromptChoice.TRADING_AGENT
     task_description = AgentIdentifier.MICROCHAIN_AGENT_OMEN
+    goal_manager: GoalManager | None = None
```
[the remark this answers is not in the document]
evangriffiths: nooo it has to be optional. some agents don't want a goal manager

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -47,6 +54,9 @@ def run(
                 prompt_handler if self.load_historical_prompt else None
             ),
         )
+
+        goal = self.goal_manager.get_goal() if self.goal_manager else None
```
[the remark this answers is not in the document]
evangriffiths: it's not though!

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -57,13 +67,28 @@ def run(
...
 
         agent.run(self.n_iterations)
 
+        if self.goal_manager:
+            goal = check_not_none(goal)
+            goal_evaluation = self.goal_manager.evaluate_goal_progress(
+                goal=goal,
+                chat_history=ChatHistory.from_list_of_dicts(agent.history),
+            )
+            self.goal_manager.save_evaluated_goal(goal=goal, evaluation=goal_evaluation)
```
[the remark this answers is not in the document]
evangriffiths: self.goal_manager might be None

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -109,3 +137,22 @@ class DeployableMicrochainModifiableSystemPromptAgent3(
 ):
     task_description = AgentIdentifier.MICROCHAIN_AGENT_OMEN_LEARNING_3
     model = SupportedModel.llama_31_instruct
+
+
+class DeployableMicrochainWithGoalManagerAgent0(DeployableMicrochainAgent):
+    task_description = AgentIdentifier.MICROCHAIN_AGENT_OMEN_WITH_GOAL_MANAGER
+    model = SupportedModel.gpt_4o
+    goal_manager = GoalManager(
```
kongzii, line 145: I think this would bring the same issue as I had with TavilyStorage. If GoalManager is initialized on import-level, then it will also be initialized in agent runs that don't use it and don't have Postgres connection string in the environment variables (CoinFlip agent, for example), and because of it, the run will fail.

I fixed it by initialising it in the `load` method https://github.com/gnosis/prediction-market-agent/pull/409/files:

Screenshot by Dropbox Capture

prediction_market_agent/agents/microchain_agent/deploy.py
evangriffiths: Done 😃

prediction_market_agent/agents/microchain_agent/deploy.py
evangriffiths: Ah good point. I've moved this into a function that's called during `run` instead of on init, but inside `DeployableMicrochainAgent.build_goal_manager` instead (which is called after the agent is built, to also solve the problem of not hardcoding the goal manager's 'capabilities')

prediction_market_agent/agents/goal_manager.py
```diff
@@ -0,0 +1,307 @@
...
+            input_variables=[
+                "high_level_description",
+                "agent_capabilities",
+                "previous_evaluated_goals",
+            ],
+            partial_variables={"format_instructions": parser.get_format_instructions()},
+        )
+        latest_evaluated_goals_str = self.evaluated_goals_to_str(latest_evaluated_goals)
+        llm = ChatOpenAI(
+            temperature=0,
```
kongzii, line 183: This might be better with super low temperature constant I introduced somewhere, if you want consistency as much as possible

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -60,6 +74,10 @@ def run(
             ),
         )
 
+        if goal_manager := self.build_goal_manager(agent=agent):
+            goal = goal_manager.get_goal()
+        agent.prompt = goal.to_prompt() if goal else None
```
kongzii, line 79: Isn't the goal undefined if the if statement above is evaluated as false?

```suggestion
            agent.prompt = goal.to_prompt() if goal else None
```

(not sure if that's the right fix 😄 )

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -109,3 +144,20 @@ class DeployableMicrochainModifiableSystemPromptAgent3(
...
+
+class DeployableMicrochainWithGoalManagerAgent0(DeployableMicrochainAgent):
+    task_description = AgentIdentifier.MICROCHAIN_AGENT_OMEN_WITH_GOAL_MANAGER
+    model = SupportedModel.gpt_4o
+    system_prompt_choice = SystemPromptChoice.TRADING_AGENT_MINIMAL
+
+    def build_goal_manager(
+        self,
+        agent: Agent,
+    ) -> GoalManager | None:
```
kongzii, line 157:
```suggestion
    ) -> GoalManager:
```

prediction_market_agent/agents/utils.py
```diff
@@ -29,6 +29,7 @@ class AgentIdentifier(str, Enum):
     MICROCHAIN_AGENT_OMEN_LEARNING_2 = "general-agent-2"
     MICROCHAIN_AGENT_OMEN_LEARNING_3 = "general-agent-3"
     MICROCHAIN_AGENT_STREAMLIT = "microchain-streamlit-app"
+    MICROCHAIN_AGENT_OMEN_WITH_GOAL_MANAGER = "general-agent-4-with-goal-manager"
```
kongzii, line 32:
```suggestion
    MICROCHAIN_AGENT_OMEN_WITH_GOAL_MANAGER = "trader-agent-4-with-goal-manager"
```

general agents above are the ones that are learning, this one (based on trader prompt) should be only trader, if that's right, can we distinguish it?

pyproject.toml
kongzii:
```suggestion
microchain-python = { git = "https://github.com/galatolofederico/microchain.git", rev = "98e601f6b7413ea48fb0b099309d686c4b10ff5c" }
```

This is the latest now 🙏

prediction_market_agent/agents/microchain_agent/deploy.py
evangriffiths: Ah eagle eye'd spot, thanks!

Since `GoalManager.get_goal` always returns a goal, it can be a little simpler:
```
        if goal_manager := self.build_goal_manager(agent=agent):
            goal = goal_manager.get_goal()
            agent.prompt = goal.to_prompt()
```

prediction_market_agent/agents/utils.py
evangriffiths: Ah true. How about `trader-agent-0-with-goal-manager`, as there are no other `trader-agent-*`s in the list?

kongzii: Sure!

prediction_market_agent/agents/goal_manager.py
evangriffiths: I still don't feel totally convinced by the "very small temperature is best for consistency" argument (I guess the same point expressed in the comment thread on the original PR https://github.com/gnosis/prediction-market-agent-tooling/pull/307/files#r1675590362).

I re-googled to see if there was any new info, and haven't found anything to convince one way or the other. In the anthropic API docs (https://docs.anthropic.com/en/api/complete) it says:
```
Defaults to 1.0. Ranges from 0.0 to 1.0. Use temperature closer to 0.0 for analytical / multiple choice, and closer to 1.0 for creative and generative tasks.

Note that even with temperature of 0.0, the results will not be fully deterministic.
```
which suggests 0.0 is more consistent than 0.000001, but still too vague to be sure!

Anyway, all that to say, if it's alright with you I'll stick with 0.0 until we know more (and we just have inconsistent code 😅)

kongzii: Sounds good. I'm now curious to do some manual experiment about this 🙈

kongzii (approved):

PR #417, opened 2024-08-22
Streamlit fixes

kongzii:
When I tried it locally, after a few predictions, now I ended up at 1.8 GB memory instead of 2.8 GB. I think 1.8 is still too much, but I wasn't able to figure it out yet. So at least something.

### DIFF
--- a/prediction_market_agent/utils.py
+++ b/prediction_market_agent/utils.py
@@ -192,7 +192,7 @@ def patch_sqlite3() -> None:
 
         sys.modules["sqlite3"] = sys.modules.pop("pysqlite3")
     except ImportError:
-        logger.warning("pysqlite3-binary not found, using sqlite3 instead.")
+        logger.debug("pysqlite3-binary not found, using sqlite3 instead.")
 
 
 def disable_crewai_telemetry() -> None:
--- a/scripts/agent_app.py
+++ b/scripts/agent_app.py
@@ -11,10 +11,12 @@ st.set_page_config(layout="wide")
 from prediction_market_agent.utils import patch_sqlite3
 
 patch_sqlite3()
+import gc
 import time
 import typing as t
 
 from prediction_market_agent_tooling.config import APIKeys
+from prediction_market_agent_tooling.markets.agent_market import AgentMarket
 from prediction_market_agent_tooling.markets.markets import (
     MarketType,
     get_binary_markets,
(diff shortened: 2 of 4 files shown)


prediction_market_agent/utils.py
```diff
@@ -192,7 +192,7 @@ def patch_sqlite3() -> None:
 
         sys.modules["sqlite3"] = sys.modules.pop("pysqlite3")
     except ImportError:
-        logger.warning("pysqlite3-binary not found, using sqlite3 instead.")
+        logger.debug("pysqlite3-binary not found, using sqlite3 instead.")
```
kongzii, line 195: People don't need to see this each time they open our Stremalit app 😄

scripts/agent_app.py
```diff
@@ -62,6 +64,45 @@
 add_sink_to_logger()
 
 
+def predict(
```
kongzii, line 67: Just cut it out into a separate function

scripts/agent_app.py
```diff
@@ -118,39 +145,35 @@ def agent_app() -> None:
...
-            # Show the agent's title.
-            st.write(
-                f"## {agent.__class__.__name__.replace('Deployable', '').replace('Agent', '')}"
-            )
-
-            with st.spinner("Agent is processing the market..."):
-                processed_market = agent.process_market(
-                    market_source, market, verify_market=not skip_market_verification
-                )
+    # Select an agent from the list of available agents.
```
kongzii, line 148: I moved this down so we can first select a question and then an agent. Otherwise, the agent starts predicting right away some random question that we didn't even want.

scripts/agent_app.py
```diff
@@ -118,39 +145,35 @@ def agent_app() -> None:
...
+                    market=market,
+                    skip_market_verification=skip_market_verification,
+                    enable_langfuse=APIKeys().default_enable_langfuse
+                    and not custom_question_input  # Don't store custom inputs, as we won't have true data for them.
+                    and not skip_market_verification,  # Don't store unverified markets, as they wouldn't be predicted anyway.
                 )
-            )
+            finally:
+                # Force memory cleanup no matter what.
+                gc.collect()
```
kongzii, line 176: And always call gc to be on the safe side

evangriffiths (approved):

PR #419, opened 2024-08-23
Save microchain agent history even if an exception is thrown

evangriffiths:
See an example where an agent runs for some number of iterations, but throws an exception:
Screenshot 2024-08-23 at 09 53 28

On https://deployed-general-agent-viewer.ai.gnosisdev.com/ the agent's chat history is empty for that session, as the agent doesn't save its history up to the point of the exception
Screenshot 2024-08-23 at 10 15 53

### DIFF
--- a/prediction_market_agent/agents/microchain_agent/deploy.py
+++ b/prediction_market_agent/agents/microchain_agent/deploy.py
@@ -64,8 +64,6 @@ class DeployableMicrochainAgent(DeployableAgent):
 
         try:
             agent.run(self.n_iterations)
-        except Exception as e:
-            raise e
         finally:
             save_agent_history(
                 agent=agent,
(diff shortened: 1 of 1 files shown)


prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -62,15 +62,18 @@ def run(
...
-        save_agent_history(
-            agent=agent,
-            long_term_memory=long_term_memory,
-            initial_system_prompt=initial_formatted_system_prompt,
-        )
-        if agent.system_prompt != initial_formatted_system_prompt:
-            prompt_handler.save_prompt(get_editable_prompt_from_agent(agent))
+        try:
+            agent.run(self.n_iterations)
+        except Exception as e:
```
kongzii, line 67: This can be omitted if you don't do anything here

```suggestion
```

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -62,15 +62,18 @@ def run(
...
+            raise e
```
kongzii, line 68:
```suggestion
```

kongzii (approved):

PR #440, opened 2024-09-02
OFVChallenger

kongzii:


prediction_market_agent/agents/ofvchallenger_agent/deploy.py
```diff
@@ -1,21 +1,132 @@
...
 from prediction_market_agent.utils import APIKeys
 
 OFV_CHALLENGER_TAG = "ofv_challenger"
+OFV_CHALLENGER_EOA_ADDRESS = Web3.to_checksum_address(
+    "0x03aEEE267A985ab9835c025377F0D8209901a928"
+)
+OFV_CHALLENGER_SAFE_ADDRESS = Web3.to_checksum_address(
+    "0x9D0260500ba7b068b5b0f4AfA9F8864eBc0B059a"
+)
+CHALLENGE_BOND = xdai_type(10)
```
kongzii, line 34: Approved by Martin, I'm going to ask finance for funds.

prediction_market_agent/agents/ofvchallenger_agent/deploy.py
```diff
@@ -1,21 +1,132 @@
...
+    claim_all_bonds_on_reality,
+)
 from prediction_market_agent.utils import APIKeys
 
 OFV_CHALLENGER_TAG = "ofv_challenger"
+OFV_CHALLENGER_EOA_ADDRESS = Web3.to_checksum_address(
+    "0x03aEEE267A985ab9835c025377F0D8209901a928"
+)
+OFV_CHALLENGER_SAFE_ADDRESS = Web3.to_checksum_address(
+    "0x9D0260500ba7b068b5b0f4AfA9F8864eBc0B059a"
```
kongzii, line 32: This agent should be quite rich, so I think it's a good one to test out our Safe implementation.

prediction_market_agent/agents/ofvchallenger_agent/deploy.py
```diff
@@ -1,21 +1,146 @@
...
+            logger.info(
+                f"Market {market.url=} already challenged with bond > {CHALLENGE_BOND} / 2. Skipping."
+            )
+            return None
+
+        # We don't plan to re-challenge markets already challenged by the challenger, should we?
+        if any(
+            response.user_checksummed == OFV_CHALLENGER_SAFE_ADDRESS
+            for response in existing_responses
+        ):
```
evangriffiths, line 111: This should already be covered by the previous check, right? Might be good to move it before it though, so you get more informative logging (this message if we've already challenged it vs other message if someone else has challenged it)

prediction_market_agent/agents/ofvchallenger_agent/deploy.py
```diff
@@ -1,21 +1,146 @@
...
+            logger.info(
+                f"Market {market.url=} already challenged by replicator. Skipping."
```
evangriffiths, line 113:
```suggestion
                f"Market {market.url=} already challenged by challenger. Skipping."
```

prediction_market_agent/agents/ofvchallenger_agent/deploy.py
```diff
@@ -1,21 +1,146 @@
...
-    def challenge(self) -> None:
+    def challenge(self, api_keys: APIKeys) -> None:
         self.langfuse_update_current_trace(tags=[OFV_CHALLENGER_TAG])
-        APIKeys()
-        # TODO: Coming in the next PR.
+
+        # Claim the bonds as first thing, to have funds for the new challenges.
+        claim_all_bonds_on_reality(api_keys)
+
+        markets_open_for_answers = OmenSubgraphHandler().get_omen_binary_markets(
```
evangriffiths, line 68: does this include markets that are resolved as invalid? if so do we want to also challenge these?

prediction_market_agent/agents/ofvchallenger_agent/deploy.py
```diff
@@ -1,21 +1,146 @@
...
+            api_keys=api_keys,
+            market=market,
+            resolution=new_resolution,
+            bond=CHALLENGE_BOND,
+            web3=web3,
+        )
+
+        return Challenge(
+            old_responses=existing_responses, new_resolution=new_resolution
+        )
```
evangriffiths, line 146: I was expecting this section to only run if the new resolution disagreed with the existing one (else return None)

prediction_market_agent/agents/ofvchallenger_agent/deploy.py
kongzii: Yeah, ideally it should be. Flipped them.

prediction_market_agent/agents/ofvchallenger_agent/deploy.py
kongzii: > does this include markets that are resolved as invalid?

Yes it should, by default we don't have any filter for invalid answers in place, only if we ask for resolved markets:

Screenshot by Dropbox Capture

>  if so do we want to also challenge these?

I think we do. If Olas market-creator resolve it as Invalid, it doesn't mean it's true (same as with Yes/No resolutions).

However, currently if OFV can not find the answer, we won't challenge it:

Screenshot by Dropbox Capture

Should we challenge to Invalid if `factuality` is `None`?

prediction_market_agent/agents/ofvchallenger_agent/deploy.py
kongzii: > I was expecting this section to only run if the new resolution disagreed with the existing one (else return None).

Martin agreed to $10 bond, so we can challenge everything to increase the bond to attract people (even if OFV and Olas agree, it doesn't mean it's correct). 

> Also, looks like the returned Challenge object isn't used anywhere (so maybe not necessary?)

It will be saved in Langfuse, which is just a practicality for debugging.

evangriffiths: > Martin agreed to $10 bond and losing 10% on 15 markets daily, so we can challenge everything to increase the bond to attract people (even if OFV and Olas agree, it doesn't mean it's correct).

Ah okay, we're taking the nuclear option!

prediction_market_agent/agents/ofvchallenger_agent/deploy.py
evangriffiths: Yes, I think that makes sense. When you ran the benchmark to compare current resolutions to OFV's, did this include any invalid markets? Wondering if it's a bad idea to do this if we don't know how good it is as evaluating invalid markets.

Similarly we could challenge as invalid if `answer is None` (i.e. `is_predictable_binary(market_question) == False`), although this isn't using any OFV stuff.

kongzii: >  did this include any invalid markets?

Unfortunatelly no.

>  Wondering if it's a bad idea to do this

I guess we can? It's a valid output from OFV and there will be bond for people to fix it 🤷 

> Similarly we could challenge as invalid if answer is None

However, I wouldn't do this. Because if `asnwer is None` it can mean arbitrary exception. in OFV run(missing credits, pydantic model problem, etc.).

> is_predictable_binary(market_question) == False

That would mean bug in Olas' market-creator, maybe we should log it as error in this agent? But I wouldn't resolve it as invalid, because non-binary markets are valid markets, just not supported by us.

evangriffiths: Good points!

> Should we challenge to Invalid if factuality is None?

Okay, let's got for it!

kongzii: Updated 🙏 YOLO style!

evangriffiths (approved):

Issue #211, opened 2024-05-22
Unify `main.py` and `prediction_market_agent/run.py`


evangriffiths:
Currently 

- Remove main.py
- Remove `prediction_market_agent/agents/all_agents.py`
- All agents from `all_agents.py` not already in `prediction_market_agent/run.py` should be added there
- Now `prediction_market_agent/run.py` is the entry point to running all agents we have
- (Optional/Contentious) add a deployment for each of these new `Deployable*Agents` (maybe just 1 markets/day), OR just remove it!

PR #446, opened 2024-09-03
Remove unused agents and scripts

evangriffiths:
PMA these days is less of an experimentation ground for new agent frameworks, and more of a place where our production agents live. Updating README and removing unused agents/scripts to reflect that.

Fixes https://github.com/gnosis/prediction-market-agent/issues/211


kongzii (approved):
Cool, Thanks!

PR #447, opened 2024-09-03
Fix mypy

evangriffiths:
Looks like PMA got out of sync with PMAT at some point

See https://github.com/gnosis/prediction-market-agent-tooling/blob/f4e226b0bc03d8cee00687083609573126e3a7d5/prediction_market_agent_tooling/markets/omen/omen_resolving.py#L39

### DIFF
--- a/prediction_market_agent/agents/replicate_to_omen_agent/omen_resolve_replicated.py
+++ b/prediction_market_agent/agents/replicate_to_omen_agent/omen_resolve_replicated.py
@@ -79,7 +79,6 @@ def omen_finalize_and_resolve_and_claim_back_all_markets_based_on_others_tx(
         api_keys,
         created_not_claimed_questions,
         auto_withdraw=True,
-        silent_errors=True,
     )
     balances_after_claiming = get_balances(public_key)
     logger.info(f"{balances_after_claiming=}")


prediction_market_agent/agents/replicate_to_omen_agent/omen_resolve_replicated.py
```diff
@@ -79,7 +79,6 @@ def omen_finalize_and_resolve_and_claim_back_all_markets_based_on_others_tx(
         api_keys,
         created_not_claimed_questions,
         auto_withdraw=True,
-        silent_errors=True,
```
kongzii, line 82: That's my change, but how come mypy didn't catch it 🤔 Anyway, thank you

kongzii (approved):

prediction_market_agent/agents/replicate_to_omen_agent/omen_resolve_replicated.py
kongzii: Oh wait, got it!

PMA is versioned to use ^0.48.6 of PMAT https://github.com/gnosis/prediction-market-agent/blob/main/pyproject.toml#L37

But this change was introduced in 0.48.8 https://github.com/gnosis/prediction-market-agent-tooling/releases

I guess you have now locally installed a new version of PMAT from your other PR?

evangriffiths: :)))))) no worries!

PR #471, opened 2024-09-16
Adjust Replication fetching

kongzii:


prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
```diff
@@ -66,7 +65,11 @@ def omen_replicate_from_tx(
         10 if market_type == MarketType.POLYMARKET else 1000,
         market_type,
         filter_by=FilterBy.OPEN,
-        sort_by=sort_by,
+        sort_by=(
+            SortBy.NONE
+            if market_type == MarketType.POLYMARKET
```
kongzii, line 70: Polymarket doesn't allow sorting by closing soonest

prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
```diff
@@ -66,7 +65,11 @@ def omen_replicate_from_tx(
...
+            else SortBy.CLOSING_SOONEST
```
kongzii, line 71: I noticed today that on Manifold, when replication form markets closing in less than a week, we have 6 markets available with default sorting or 460 markets with this sorting. So this gives more opportunities to actually go through our filters (is_predictable, is_predictable_without_description) and replicate something.

prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
gabrielfior: I would suggest sorting by Volume - this way, we have the "most interesting" (or relevant) markets.

Maybe for another ticket - when creating the market, can we create it with different odds than the default 50/50? I would say we should copy the odds from the source market.

kongzii: It's further sorted by volume just below this: 

Screenshot by Dropbox Capture

> Maybe for another ticket - when creating the market, can we create it with different odds than the default 50/50? I would say we should copy the odds from the source market.

Yes that's possible

gabrielfior: @coderabbit create a new issue out of this discussion

gabrielfior: Referenced question 2 [here](https://github.com/gnosis/prediction-market-agent-tooling/issues/406). Closing this.

gabrielfior (approved):

PR #475, opened 2024-09-19
Update some agents to bet on 20 markets per run

kongzii:


prediction_market_agent/agents/known_outcome_agent/deploy.py
```diff
@@ -16,6 +16,7 @@
 
 
 class DeployableKnownOutcomeAgent(DeployableTraderAgent):
+    bet_on_n_markets_per_run = 20
     model = "gpt-4-1106-preview"
```
kongzii, line 20: @evangriffiths gpt-4 is the most expensive, but it's known outcome that's a most profitable, hard choice, wdyt?

prediction_market_agent/agents/prophet_agent/deploy.py
```diff
@@ -75,6 +76,7 @@ def load(self) -> None:
 
 class DeployableOlasEmbeddingOAAgent(DeployableTraderAgentER):
     agent: OlasAgent
+    bet_on_n_markets_per_run = 20
```
kongzii, line 79:
```
input $2.5 / 1M tokens
output $10.00 / 1M tokens
```

prediction_market_agent/agents/prophet_agent/deploy.py
```diff
@@ -99,6 +101,7 @@ def load(self) -> None:
 
 
 class DeployablePredictionProphetGPTo1MiniAgent(DeployableTraderAgentER):
+    bet_on_n_markets_per_run = 20
```
kongzii, line 104: input $3 / 1M tokens
output $12.00 / 1M tokens

kongzii:
Note: If the current agent runs every 2 or 3 hours, 12 or 8 markets per day. After this change, it's 160 or 240 markets per day.

@evangriffiths @gabrielfior Are you sure it should be 20? Worst case we run out of credits but FYI to keep it in mind.

prediction_market_agent/agents/prophet_agent/deploy.py
evangriffiths: maybe a side-issue, but I wonder if we should retire this guy. Not sure what benefit we have from keeping it deployed - we track the actual olas agents on dune, and the code its using to make predictions is from the prediction_prophet repo, so over time it will get out-of-date with what the olas agents are doing anyway.

kongzii: i thought we can just leave him as long as he doesn't die out, but the name is not great that's true

PR #482, opened 2024-09-20
Fix parallelism in Thinker

kongzii:


prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -5,15 +5,16 @@
 
 import tenacity
 from crewai import Agent, Crew, Process, Task
-from langchain.utilities.tavily_search import TavilySearchAPIWrapper
 from langchain_community.tools.tavily_search import TavilySearchResults
+from langchain_community.utilities.tavily_search import TavilySearchAPIWrapper
```
kongzii, line 9: Just different import location to get rid of warning logs from langchain

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -5,15 +5,16 @@
...
 from langchain_core.callbacks import (
     AsyncCallbackManagerForToolRun,
     CallbackManagerForToolRun,
 )
 from langchain_core.language_models import BaseChatModel
 from langchain_core.pydantic_v1 import SecretStr
 from langchain_openai import ChatOpenAI
+from loky import get_reusable_executor
```
kongzii, line 17: Use this instead of the previously used function from PMAT https://github.com/joblib/loky

The previous implementation works just fine locally and initially worked in Kubernetes, too. I don't know what changed, but this seems to fix it.

They are more fancy:

Screenshot by Dropbox Capture

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -5,8 +5,8 @@
 
 import tenacity
 from crewai import Agent, Crew, Process, Task
-from langchain.utilities.tavily_search import TavilySearchAPIWrapper
 from langchain_community.tools.tavily_search import TavilySearchResults
+from langchain_community.utilities.tavily_search import TavilySearchAPIWrapper
```
kongzii, line 9: Fixing import location based on warning message

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
```diff
@@ -257,7 +253,7 @@ def get_correlated_markets(self, question: str) -> list[CorrelatedMarketInput]:
 
         markets = par_map(
             items=[q.market_address for q in nearest_questions],
-            func=lambda market_address: self.subgraph_handler.get_omen_market_by_market_id(
+            func=lambda market_address: OmenSubgraphHandler().get_omen_market_by_market_id(
```
kongzii, line 256: `self` can't be pickled into Loky

gabrielfior: Can the `OmenSubgraphHandler` be instantiated outside of the lambda func?
It costs a bit (~1-3s) to load everything, and the instance can be shared since requests can be sent in parallel to the subgraph.
This can speed up the processing.

kongzii: Didn't you make it a singleton that's just a single instance? https://github.com/gnosis/prediction-market-agent-tooling/blob/45601b68c827206a85dcf194732c1f3990fd7509/prediction_market_agent_tooling/markets/omen/omen_subgraph_handler.py#L50

(btw it won't be shared, but copied over, because these are individual processes, but that doesn't change your argument, just a terminology neatpick)

gabrielfior: image

Actually `__init__` is being called every time the `SubgraphHandler` is instantiated. So this means the `Singleton` implementation is buggy.
Would you agree to call the instance outside of the lambda function and create a new ticket for fixing the singleton>?

kongzii: > Create a new ticket for fixing the singleton>?

Yep that's needed. @coderabbitai Please create a ticket for that.

> Would you agree to call the instance outside of the lambda function

Why? We need to fix it, and then it will be fine (we are already assuming at many places in code that it's a singleton that isn't re-initialized, so it needs a global fix, not local)

gabrielfior: The current implementation of the Singleton needs a fix - so your current usage will require more time to complete until the fix is done.
If you are fine with that, leave as is.
If not, consider moving the instantiation outside of the loop.

kongzii: > If you are fine with that, leave as is.

I am, we are talking about saving 1-3 seconds in a computation that takes 1 to 5 minutes

gabrielfior (approved):

prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py
gabrielfior: Excellent. Approved.

PR #502, opened 2024-10-08
Bet on more markets

kongzii:
It seems like we are doing fine on OpenAI (50% of the budget used last month) and Tavily (it doesn't show last month's usage, but currently on 7k out of 38k credits).

Shall we bet a bit more?


evangriffiths (approved):
💥

PR #506, opened 2024-10-09
Use is_invalid in OFV

kongzii:


evangriffiths (approved):

prediction_market_agent/agents/ofvchallenger_agent/ofv_resolver.py
```diff
@@ -101,9 +102,21 @@ def ofv_answer_binary_question(
...
+    question_is_invalid = is_invalid(market_question)
+    if question_is_invalid:
+        logger.warning(
+            f"Question `{market_question}` is invalid, skipping fact checking and returning invalid resolution."
+        )
+        return FactCheckAnswer(
+            factuality=None,
+            chosen_results=[],
+            all_considered_results=[],
+        )
```
gabrielfior, line 115: If I understand correctly, 'factuality=None' triggers OFVChallengerAgent::challenge_market::omen_submit_invalid_answer_market_tx`, i.e. it will answer the market as invalid.
To me, `factuality` in general is not clear - can we somehow handle the Outcomes we are picking as Answers, for ex, returning Resolution as a property? That would avoid some `if else` logic inside `challenge_market`.

kongzii: `factuality` is what OFV is returning internally, so the if-else conversion into Resolution needs to be done somewhere by us.

Anyway this PR just fixes the logic, if you have an idea how to refactor feel free to propose something.

PR #509, opened 2024-10-10
New Metaculus Q4 tournament

kongzii:


kongzii:
As per Evan's comment on Slack, currently blocked by:

```
I think yes, but also:
do this first https://github.com/gnosis/prediction-market-agent/issues/371 (add market resolution criteria to agent prompt)
use our current best performing agent (which according to dune, is think-thoroughly-prophet (currently uses ThinkThoroughlyWithItsOwnResearch)
```

evangriffiths (approved):

PR #510, opened 2024-10-10
Fix challenging agents (replicator, challenger)

kongzii:


prediction_market_agent/agents/ofvchallenger_agent/deploy.py
```diff
@@ -68,14 +69,21 @@ def challenge(self, api_keys: APIKeys) -> None:
...
+        )
+        markets_open_for_answers = get_omen_binary_markets_common_filters(
             # With a little bandwidth for the market to be finalized,
             # so we have time for processing it without erroring out at the end.
-            finalized_after=utcnow() + timedelta(minutes=30),
+            question_finalized_after=utcnow()
+            + timedelta(minutes=30),
+        ) + get_omen_binary_markets_common_filters(
+            # And also markets without any answer at all yet.
+            question_with_answers=False,
```
kongzii, line 86: I didn't realise before that if I use `question_finalized_after`, it won't include questions without any answer at all (where it's `null`)

gabrielfior: Effectivelly what you are doing with the `partial` is an OR filter, since you want (in this code snippet) the questions that finalize after `now() + 30min` OR questions without answers.
For this, I find querying the subgraph using `OR` a lot cleaner - see code snippet below.
```
{
  questions(first: 5,
  where: {
    or: [{
    id: "0x001dd5a9194948a903c0e3b624eb0d458cd23d828f29a57f517c858d7cf71f76"  
    },
    {id : "0x0005ef01269fbb37aba4cc16f518138d7d8bf5f4937a5572c294c2e62f0cf670"}
    ],    
  }) {
    id
    templateId
    data
    title
  }
}
```
What do you think?

kongzii: > using OR a lot cleaner

It is clean in the simple example query you provided, but how do you suggest integrating `or` functionality into OmenSubgraphHandler? With any integration that I can imagine, it's in the end easier to just do it like I do it here.

- `finalized_after` shouldn't be `finalized_after OR without_answers`, because sometimes we might need to really query only `finalized_after` markets
- I didn't want to introduce `finalize_after_or_null` filter, because that we would have to have `or_null` variant of many other arguments as well

gabrielfior: I agree `partial` here is simple and works, but what I was thinking is something along these lines (inside the subgraph handler) if the use case becomes more complicated:
```
where_stm1 = -> build with params 1
where_stm2 -> build with params2
where_final = {"or": [where_stms1, where_stms2]}
```
The added benefit here is to execute just 1 query, whereas in yours we execute 2 queries.

gabrielfior: But happy here with partial :)

gabrielfior (approved):

prediction_market_agent/agents/ofvchallenger_agent/deploy.py
kongzii: That will need some fancy refactoring of OmenSubgraphHandler, but I agree it could be nice!

I'm not sure how it would end up in the end, but if you are up to exploring it, it sounds good to me.

> to execute just 1 query, whereas in yours we execute 2 queries.

That's a good benefit out of it.

PR #511, opened 2024-10-10
137 arbitrage agent

gabrielfior:
-> Added Arbitrage agent for placing YES/NO or NO/YES trades between correlated markets
-> Added test

---
Some examples of correlated markets + expected profit
```
"correlated_markets[0].correlation=1.0 
correlated_markets[0].main_market.question='Will the average global temperature in 2024 exceed 2023?' correlated_markets[0].related_market.question='Will 2024 be even hotter than 2023? 🥵 '"
 potential_profit = 0.06270421212134791

- [main_market Will Donald Trump be the Republican nominee for president in 2024? related_market_question Will Donald Trump Be The GOP Candidate for The 2024 Presidential Election potential profit 0.23373115547666723]
- [main_market Will the LK-99 room temp, ambient pressure superconductivity pre-print replicate before 2025? related_market_question Will the Meissner effect be confirmed near room temperature in copper-substituted lead apatite? potential profit 9.799516753616899e-11]
- [main_market Will Donald Trump win the 2024 presidential election? related_market_question Will Donald Trump win the 2024 US Presidential Election? potential profit 0.004095501879286534, main_market Will Donald Trump win the 2024 presidential election? related_market_question Will Trump win the 2024 US election and be elected as president? potential profit 0.42199872849628406]
- [main_market Will Netanyahu remain Israel's PM until the end of 2024? related_market_question [ACX 2024] Will Benjamin Netanyahu remain Prime Minister of Israel throughout 2024? potential profit 0.0, main_market Will Netanyahu remain Israel's PM until the end of 2024? related_market_question Will Netanyahu still be the prime minister of Israel at the end of October 7th 2024? potential profit 0.30326927261165015, main_market Will Netanyahu remain Israel's PM until the end of 2024? related_market_question Will Benjamin Netanyahu (Bibi) be the prime minister of Israel at the end of 2024 potential profit 0.0]
```

Edit - now that correlation < 0 (`near_perfect_correlation = False`) can also be handled, found some inversely correlated pairs
```
Correlation(near_perfect_correlation=False, reasoning="The two markets are inversely correlated. MARKET 1 asks if Netanyahu will remain Israel's PM until the end of 2024, while MARKET 2 asks if Netanyahu will die before October 7th, 2024. If MARKET 2 resolves to 'Yes' (Netanyahu dies before October 7th, 2024), then MARKET 1 must resolve to 'No' (he cannot remain PM until the end of 2024 if he dies before then). Conversely, if MARKET 1 resolves to 'Yes', then MARKET 2 must resolve to 'No'. Therefore, the outcomes of these markets are perfectly inversely correlated.")
```


prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,193 @@
...
+    ) -> list[Trade]:
+        trades = []
+        correlated_markets = self.get_correlated_markets(market=market)
+        # For each correlated market, we want to place YES/NO or NO/YES trades.
+        for pair in correlated_markets:
+            if pair.potential_profit_per_bet_unit > 0:
+                trades_for_pair = self.build_trades_for_correlated_markets(pair)
+                trades.extend(trades_for_pair)
+
+        return trades
```
gabrielfior, line 193: I found it easier to build out the logic inside `build_trades` than to implement a whole new strategy, which would not follow the current interface on `BettingStrategy`.

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,193 @@
...
+from prediction_market_agent.utils import APIKeys
+
+
+class DeployableOmenArbitrageAgent(DeployableTraderAgent):
+    """Agent that places mirror bets on Omen for (quasi) risk-neutral profit."""
+
+    correlation_threshold: float = 0.8
+
+    def load(self) -> None:
+        self.subgraph_handler = OmenSubgraphHandler()
```
evangriffiths, line 49: Any reason you've made this agent omen-specific? Looks like the only thing you've used the sgh for is to get markets. So it looks like this could easily be changed to become market-platform agnostic

evangriffiths (commented):
Could you share some examples of real-world correlated markets found by this agent (+ the trades it makes + the expected profit) in the PR description, so we can eyeball that it looks like the agent will work when released in the wild?

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,193 @@
...
+    Correlation,
+)
+from prediction_market_agent.db.pinecone_handler import PineconeHandler
+from prediction_market_agent.utils import APIKeys
+
+
+class DeployableOmenArbitrageAgent(DeployableTraderAgent):
+    """Agent that places mirror bets on Omen for (quasi) risk-neutral profit."""
+
+    correlation_threshold: float = 0.8
```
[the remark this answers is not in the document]
gabrielfior: I also like (2) better, but markets that are obviously strongly correlated, e.g. like in the test
```
 m1.question = "Will Kamala Harris win the US presidential election?"
m1.question = "Will Kamala Harris become the US president in 2025?"
```
only deliver 0.8 correlation.

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,193 @@
...
+                    amount=amount_yes, currency=market_to_bet_yes.currency
+                ),
+            ),
+            Trade(
+                trade_type=TradeType.BUY,
+                outcome=False,
+                amount=TokenAmount(
+                    amount=amount_no, currency=market_to_bet_no.currency
+                ),
+            ),
```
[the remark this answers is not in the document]
gabrielfior: Positive correlation is assumed here -> https://github.com/gnosis/prediction-market-agent/pull/511/files/433656d955694050935c662e1574f80c552d3cc7#diff-1b9d8df9eb7cb55918e642ebc0f26902d5211d4946a8c4d0454e9889dd05c371R138-R140

I didn't implement logic for the negatively correlated case (it was on an earlier version), I think positive correlation is a good start. Also, negative correlation is harder to determine, e.g. `Trump president vs Kamala president` is 100% negatively correlated but not trivial for LLM to tell, hence preferred to stick to 100% correlation case.

Happy with suggestions though.

prediction_market_agent/agents/arbitrage_agent/deploy.py
gabrielfior: I made it Omen-specific because the `related_markets` should be queryable via Pinecone (because of the TTA agent), and only Omen agents are stored there currently.

tests/agents/arbitrage_agent/test_arbitrage_agent.py
```diff
@@ -0,0 +1,66 @@
...
+def main_market() -> t.Generator[AgentMarket, None, None]:
+    m1 = Mock(OmenAgentMarket, wraps=OmenAgentMarket)
+    m1.question = "Will Kamala Harris win the US presidential election?"
+    yield m1
+
+
+@pytest.fixture(scope="module")
+def related_market() -> t.Generator[AgentMarket, None, None]:
+    m1 = Mock(OmenAgentMarket, wraps=OmenAgentMarket)
+    m1.question = "Will Kamala Harris become the US president in 2025?"
```
evangriffiths, line 37: I'm a bit confused by this, if you expect high correlation with the other question shouldn't this be either:
- `"Will Kamala Harris be the US president in 2025?"` or
- `"Will Kamala Harris become the US president in 2024?"`

prediction_market_agent/agents/arbitrage_agent/deploy.py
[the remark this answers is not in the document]
evangriffiths: Oh okay. Fwiw I think it's okay if the LLMs find it hard to determine that, as long as there aren't false positives.

tests/agents/arbitrage_agent/test_arbitrage_agent.py
evangriffiths: Oh scrap that sorry, elected president takes office in Jan 2025. Forgot!

prediction_market_agent/agents/arbitrage_agent/deploy.py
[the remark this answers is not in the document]
evangriffiths: I ran it with gpt-4o instead of 4o-mini and it gives 1.0!!

```
from prediction_market_agent.agents.arbitrage_agent.deploy import (
    DeployableOmenArbitrageAgent,
)

agent = DeployableOmenArbitrageAgent()
correlation = agent.chain.invoke(
    {
        "main_market_question": "Will Kamala Harris win the US presidential election?",
        "related_market_question": "Will Kamala Harris become the US president in 2025?",
    }
)
print(correlation)
```

Maybe use that instead? Should be very cheap to run these prompts. Have you checked?

prediction_market_agent/agents/arbitrage_agent/data_models.py
```diff
@@ -0,0 +1,70 @@
+import typing as t
+
+from prediction_market_agent_tooling.markets.agent_market import AgentMarket
+from pydantic import BaseModel, computed_field
+
+
+class Correlation(BaseModel):
+    correlation: float
```
evangriffiths, line 8: would be good to add a `reasoning` field so we can debug from the langfuse traces if it's not behaving as expected.

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,193 @@
...
+
+                    [MARKET 1]
+                    {main_market_question}
+
+                    [MARKET 2]
+                    {related_market_question}
+
+                    Follow the formatting instructions below for producing an output in the correct format.
+                    {format_instructions}
+                    """
```
evangriffiths, line 89: indentation seems a bit funny. Should be flush to left hand side for no unecessary whitespace in the prompt 🤑.

prediction_market_agent/agents/arbitrage_agent/deploy.py
gabrielfior: Indexing occurs below
https://github.com/gnosis/prediction-market-agent/blob/a1931b8be64cbc858a139059279b971485064358/prediction_market_agent/agents/think_thoroughly_agent/think_thoroughly_agent.py#L190

I don't think it prevents someone from running the agent - you just need Pinecone api keys and, on the first run, it will fill out Pinecone and have the same dataset as we do.

> Regardless, there doesn't seem to be a need to have any Omen dependencies here - you could just have a if market_type != MarketType.Omen: raise ... check

Makes sense to me, implementing this.

prediction_market_agent/agents/arbitrage_agent/deploy.py
gabrielfior: Removed.

prediction_market_agent/agents/arbitrage_agent/deploy.py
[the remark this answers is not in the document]
gabrielfior: Sounds like a good approach, implemented

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,186 @@
...
+from prediction_market_agent.agents.arbitrage_agent.data_models import (
+    CorrelatedMarketPair,
+    Correlation,
+)
+from prediction_market_agent.agents.arbitrage_agent.prompt import prompt_template
+from prediction_market_agent.db.pinecone_handler import PineconeHandler
+from prediction_market_agent.utils import APIKeys
+
+
+class DeployableArbitrageAgent(DeployableTraderAgent):
```
evangriffiths, line 45: Any reason you're subclassing `DeployableTraderAgent` instead of `DeployableAgent`? Not sure which is better, just asking

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,193 @@
...
+                    )
+                )
+        return correlated_markets
+
+    @observe()
+    def build_trades_for_correlated_markets(
+        self, pair: CorrelatedMarketPair
+    ) -> list[Trade]:
+        market_to_bet_yes, market_to_bet_no = pair.main_market, pair.related_market
+        total_amount: BetAmount = pair.main_market.get_tiny_bet_amount()
```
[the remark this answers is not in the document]
gabrielfior: Implemented

prediction_market_agent/agents/arbitrage_agent/deploy.py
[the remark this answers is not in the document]
gabrielfior: Going with your `bool` suggestion from the other thread.

prediction_market_agent/agents/arbitrage_agent/deploy.py
gabrielfior: I'm doing so because other trading agents (`KnownOutcome`, `TTA` and even `coinflip`) are subclassing the `DeployableTraderAgent` - as I understand,it does some langfuse magic and checks for some balances, which doesn't hurt.

prediction_market_agent/agents/arbitrage_agent/deploy.py
evangriffiths: Well it does if the user wants to run just this agent, and not the DeployableThinkThoroughlyAgent as well! I think we need to do `self.pinecone_handler.insert_all_omen_markets_if_not_exists(...)` somewhere here

evangriffiths: Just had a look at what `pinecone_handler.update_markets` is doing. Is this what we want? Two things I'm not sure about:
1. It doesn't look like it's removing closed markets from the index
2. It only fetches the last 7 days' worth of created markets `created_after = utcnow() - datetime.timedelta(days=7)`. Don't we want all open markets?

prediction_market_agent/agents/arbitrage_agent/data_models.py
```diff
@@ -0,0 +1,73 @@
...
+class Correlation(BaseModel):
+    near_perfect_correlation: bool
+    reasoning: str
+
+
+class CorrelatedMarketPair(BaseModel):
+    main_market: AgentMarket
+    related_market: AgentMarket
+
+    def __repr__(self) -> str:
```
kongzii, line 16:
```suggestion
    def __str__(self) -> str:
```

Reason:

Screenshot by Dropbox Capture

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,187 @@
...
+    def _build_chain(self) -> RunnableSerializable[t.Any, t.Any]:
+        llm = ChatOpenAI(
+            temperature=0,
+            model=self.model,
+            api_key=APIKeys().openai_api_key_secretstr_v1,
+        )
+
+        parser = PydanticOutputParser(pydantic_object=Correlation)
+        prompt = PromptTemplate(
+            template=prompt_template,
```
kongzii, line 91:
```suggestion
            template=PROPT_TEMPLATE,
```

I was just looking where is this var coming from 🙈 

Screenshot by Dropbox Capture

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,187 @@
...
+    def get_correlated_markets(self, market: AgentMarket) -> list[CorrelatedMarketPair]:
+        # We try to find similar, open markets which point to the same outcome.
+        correlated_markets = []
+        related = self.pinecone_handler.find_nearest_questions_with_threshold(
+            limit=10, text=market.question
+        )
+
+        omen_markets = self.subgraph_handler.get_omen_binary_markets(
+            limit=len(related),
+            id_in=[i.market_address.lower() for i in related],
```
kongzii, line 121: Oops, can you move the lowering to PMAT? To prevent bugs in the future.

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,187 @@
...
+            resolved=False,
+        )
+        omen_markets = [m for m in omen_markets if m.id != market.id]
+        # Note that negative correlation is hard - e.g. for the US presidential election, markets on each candidate are not seen as -100% correlated.
```
kongzii, line 125: Why is this hard? If I try OpenAI platform, then without any prompting magic:

Screenshot by Dropbox Capture

prediction_market_agent/agents/arbitrage_agent/deploy.py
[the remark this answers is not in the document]
kongzii: > Also, negative correlation is harder to determine, e.g. Trump president vs Kamala president is 100% negatively correlated but not trivial for LLM to tell, hence preferred to stick to 100% correlation case.

What did you try? It seems to work fine for me, see the comment above.

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,187 @@
...
+        return super().get_markets(
+            market_type=market_type,
+            limit=50,
+            sort_by=SortBy.HIGHEST_LIQUIDITY,
+            # Fetching most liquid markets since more likely they will have related markets
+            filter_by=FilterBy.OPEN,
+        )
+
+    def answer_binary_market(self, market: AgentMarket) -> ProbabilisticAnswer | None:
+        return ProbabilisticAnswer(p_yes=Probability(0.5), confidence=1.0)
```
kongzii, line 80: How is this agent supposed to work? It just returns a fixed 50% probability for all markets.

Do you have some examples where it successfully placed bets on Omen?

Screenshot by Dropbox Capture

tests/agents/arbitrage_agent/test_arbitrage_agent.py
```diff
@@ -0,0 +1,63 @@
...
+@pytest.fixture(scope="module")
+def arbitrage_agent() -> t.Generator[DeployableArbitrageAgent, None, None]:
+    with patch(
+        "prediction_market_agent.agents.arbitrage_agent.deploy.DeployableArbitrageAgent.load",
+        new=lambda x: None,
+    ), patch(
+        "prediction_market_agent_tooling.tools.langfuse_.get_langfuse_langchain_config"
+    ):
+        agent = DeployableArbitrageAgent()
+        # needed since load was mocked
```
kongzii, line 22: I didn't see you did anything in the `load` method?

But I agree that the stuff that's done in overloaded `run` method were meant to be done in the `load` method.

tests/agents/arbitrage_agent/test_arbitrage_agent.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@pytest.fixture(scope="module")
+def unrelated_market() -> t.Generator[AgentMarket, None, None]:
+    m1 = Mock(OmenAgentMarket, wraps=OmenAgentMarket)
+    m1.question = "Will Donald Duck ever retire from his adventures in Duckburg?"
+    yield m1
+
+
+@pytest.mark.parametrize(
```
kongzii, line 48:
```suggestion
@pytest.mark.skipif(not RUN_PAID_TESTS, reason="This test costs money to run.")
@pytest.mark.parametrize(
```

tests/agents/arbitrage_agent/test_arbitrage_agent.py
gabrielfior: I discussed about this somewhere with @evangriffiths - this agent is Omen-specific, and the "is this Omen else Raise" is done on the run method (whereas `load` is called during Init, which in turn is called in `run_agent.py` before market_type is passed as arg). Hence the `self.attributes` being defined here.

prediction_market_agent/agents/arbitrage_agent/deploy.py
gabrielfior: Sure see https://github.com/gnosis/prediction-market-agent-tooling/pull/494

tests/agents/arbitrage_agent/test_arbitrage_agent.py
kongzii: Sure, sounds as a good reason to me to not use `load` method, but then this comment is incorrect.

prediction_market_agent/agents/arbitrage_agent/deploy.py
gabrielfior: Implemented changes as suggested by @evangriffiths :
-> Removing closing markets (see pinecone_handler::remove_resolved_markets)
-> Fetching all open markets instead of markets created in the last 7 days (pinecone_handler::insert_all_omen_markets_if_not_exists)

prediction_market_agent/db/pinecone_handler.py
```diff
@@ -109,15 +111,37 @@ def deduplicate_markets(markets: list[OmenMarket]) -> list[OmenMarket]:
...
+            limit=sys.maxsize,
+            resolved=True,
+        )
+        resolved_markets_id = [
+            self.encode_text(m.question_title)
+            for m in resolved_markets
+            if m.is_resolved
+        ]
+        ids_to_delete = set(existing_ids).intersection(resolved_markets_id)
+        self.remove_markets_by_id(ids_to_delete=list(ids_to_delete))
```
kongzii, line 134: Pinecone is used by both ThinkThoroughly and Arbitrage Agent. Closed markets are useful for ThinkThoroughly but unusable for Arbitrage. 

Won't this delete the closed markets from the database itself? E.g. also for TT agent?

prediction_market_agent/agents/arbitrage_agent/deploy.py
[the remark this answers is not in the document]
gabrielfior: I believe I addressed this case (negative correlation) with my latest changes, please see https://github.com/gnosis/prediction-market-agent/commit/7afbb41c3866c426ca455bfd4a655aa04610fdc7#diff-7b9692129117b2cc9d653a28cdd7bb8f5fbaf3cc6020d6ed791b8642efc63467R62-R65)

prediction_market_agent/agents/arbitrage_agent/deploy.py
gabrielfior: Closing this since it was addressed here https://github.com/gnosis/prediction-market-agent/pull/511#discussion_r1801702249

prediction_market_agent/agents/arbitrage_agent/deploy.py
gabrielfior: @kongzii as disussed during huddle today, current implementation is not awesome but works.
We agree on a new implementation (to be done https://github.com/gnosis/prediction-market-agent/issues/521)

prediction_market_agent/db/pinecone_handler.py
gabrielfior: As per our huddle, removed deletion of markets, might be useful to TTA. Closing this.

prediction_market_agent/agents/think_thoroughly_agent/deploy.py
```diff
@@ -30,7 +30,7 @@ def answer_binary_market(self, market: AgentMarket) -> ProbabilisticAnswer | Non
         )
 
     def before_process_markets(self, market_type: MarketType) -> None:
-        self.agent.update_markets()
+        self.agent.pinecone_handler.update_markets()
```
evangriffiths, line 33: according to https://github.com/gnosis/prediction-market-agent/pull/511/files#r1800902856 you're not intending to remove the closed markets from the pinecone db for the TTA. But you are changing the behaviour to only add open markets to the db (because `PineconeHandler.insert_open_omen_markets_if_not_exists` is called for both. Is this intended? how come?

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,205 @@
...
+    ProbabilisticAnswer,
+    TokenAmount,
+    Trade,
+    TradeType,
+)
+from prediction_market_agent_tooling.markets.markets import MarketType
+from prediction_market_agent_tooling.markets.omen.omen import OmenAgentMarket
+from prediction_market_agent_tooling.markets.omen.omen_subgraph_handler import (
+    OmenSubgraphHandler,
+)
```
evangriffiths, line 30: As mentioned before, would be preferable IMO to not have any omen-specific dependencies. Should be easy enough to remove now!

prediction_market_agent/agents/think_thoroughly_agent/deploy.py
gabrielfior: My understanding is this
-> We have closed (old) markets in the DB, since PineconeHandler has been running for a while - hence we should have all markets (up to the very fresh ones) in the DB
-> Now the current implementation inserts only open markets (if ID not in DB) - this makes sense because the closed markets are already in the DB, only fresh (and new ones) will be missing.
-> In the logic above, we could miss some short-lived markets, that are open-closed in the timespan of 2h (between agent runs). But I don't think these markets are very relevant.
-> Important thing here is to have the market ID (and its question title) on Pinecone for `similarity_search` - market liquidity and outcome token prices come from the subgraph (for a given market ID).

@evangriffiths does the above clear your concerns, or do you suggest a different implementation?

kongzii: > -> In the logic above, we could miss some short-lived markets, that are open-closed in the timespan of 2h (between agent runs). But I don't think these markets are very relevant.

It's a strong and already a few times disproven assumption that our agent will run every 2 hours or that we won't need to turn them off even for the whole weekend or more. 🙈 

Screenshot by Dropbox Capture

Previous behavior (fetching all of them but having a reasonable limit for created_after to not fetch thousands of markets every time) seemed fine to me.

I see Evan was worried about 

> It only fetches the last 7 days' worth of created markets created_after = utcnow() - datetime.timedelta(days=7). Don't we want all open markets?

But there should be older markets in the DB as well right? Updated from previous runs. It's cumulating in DB.

On the other hand... both implementation are wrong only in edge-cases so I guess I don't care in the end... okay my 2 cents are finished 😄

evangriffiths: What if we have to nuke our pinecone db for some reason, so we have to start from scratch - we'd want to pull in all old markets right?

prediction_market_agent/agents/arbitrage_agent/data_models.py
```diff
@@ -0,0 +1,93 @@
...
+    def bet_directions(self) -> t.Tuple[bool, bool]:
+        correlation = check_not_none(self.correlation.near_perfect_correlation)
+        if correlation:
+            # We compare denominators for cases YES/NO and NO/YES bets and take the most profitable (i.e. where denominator is the lowest).
+            # For other cases we employ similar logic.
+            yes_no = self.main_market.current_p_yes + self.related_market.current_p_no
+            no_yes = self.main_market.current_p_no + self.related_market.current_p_yes
+            return (True, False) if yes_no <= no_yes else (False, True)
+
+        else:
```
kongzii, line 59: Could you explain to me, why if `near_perfect_correlation == False`, why bet `Yes, Yes` or `No, No`? If it's not perfectly correlated (either positively or negatively), shouldn't it be skipped by this agent?

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,211 @@
...
+    def get_markets(
+        self,
+        market_type: MarketType,
+        limit: int = MAX_AVAILABLE_MARKETS,
+        sort_by: SortBy = SortBy.CLOSING_SOONEST,
+        filter_by: FilterBy = FilterBy.OPEN,
+    ) -> t.Sequence[AgentMarket]:
+        return super().get_markets(
+            market_type=market_type,
+            limit=50,
```
kongzii, line 75: `get_markets` accepts the `limit` argument, but it's overruled by this. Not a nice design.

```suggestion
            limit=limit,
```

prediction_market_agent/db/pinecone_handler.py
```diff
@@ -109,15 +112,22 @@ def deduplicate_markets(markets: list[OmenMarket]) -> list[OmenMarket]:
 
         return list(unique_market_titles.values())
 
-    def insert_all_omen_markets_if_not_exists(
-        self, created_after: DatetimeUTC | None = None
+    def update_markets(self) -> None:
```
evangriffiths, line 115: - Can remove this function as it just wraps `insert_open_omen_markets_if_not_exists`
- `insert_all_omen_markets_if_not_exists` has been changed to `insert_open_omen_markets_if_not_exists`, but it's (nearly) the same as the old function (i.e. not getting open markets only). So no need for rename.
- `created_after: DatetimeUTC` has been changed to `start_timestamp: int`. But `OmenSubgraphHandler.get_omen_binary_markets_simple(created_after=...)` takes `DatetimeUTC`. So should change back

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,211 @@
...
+            sort_by=SortBy.HIGHEST_LIQUIDITY,
```
kongzii, line 76: Same here.

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,211 @@
...
+            # Fetching most liquid markets since more likely they will have related markets
+            filter_by=FilterBy.OPEN,
```
kongzii, line 78: And here.

prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,211 @@
...
+            id_in=[i.market_address for i in related],
+            resolved=False,
+        )
+
+        # Order omen_markets in the same order as related
+        related_market_addresses = [i.market_address for i in related]
+        omen_markets = sorted(
+            omen_markets, key=lambda m: related_market_addresses.index(m.id)
+        )
+        omen_markets = [m for m in omen_markets if m.id != market.id]
```
kongzii, line 142: This can be done directly in `id_in=[i.market_address for i in related]` above.

prediction_market_agent/db/pinecone_handler.py
```diff
@@ -109,15 +112,22 @@ def deduplicate_markets(markets: list[OmenMarket]) -> list[OmenMarket]:
...
+        self.insert_open_omen_markets_if_not_exists()
+
+    def insert_open_omen_markets_if_not_exists(
+        self, start_timestamp: int | None = None
     ) -> None:
         subgraph_handler = OmenSubgraphHandler()
         markets = subgraph_handler.get_omen_binary_markets_simple(
             limit=sys.maxsize,
             filter_by=FilterBy.NONE,
+            created_after=DatetimeUTC.fromtimestamp(start_timestamp)
```
kongzii, line 127:
```suggestion
            created_after=DatetimeUTC.to_datetime_utc(start_timestamp)
```

`to_datetime_utc` is our method that is doing some more fancy stuff, check it out there 🙏

kongzii (commented):
We are the builders of PMAT, if you see something doesn't add up (like the need for returning the dummy prediction answer, or the need for over-ruling the arguments passed into `get_markets` function, let's think about how to design PMAT better, not hacking it around 🙏

prediction_market_agent/agents/arbitrage_agent/data_models.py
gabrielfior: Sure. In the case of -1 correlation, let's assume we have these markets:
-> Will Kamala be elected president?
-> Will Trump be elected president?

We should EITHER bet a. YES on both OR b. NO on both. Since correlation = -1, we will ALWAYS win exactly one of the bets.

If the above is correct, the question is only where is it "cheaper" to place the bet - if both markets are 50-50 then there is no difference, but for unbalanced markets there is a formula that dictates if bets should be (YES, YES) or (NO, NO). Maybe you can try deriving for yourself, if you have problems, happy to share my notes.

prediction_market_agent/agents/arbitrage_agent/deploy.py
gabrielfior: I understand your concern (expressed in 4 separate comments) - 50 is not limit, we should use limit.
So what's the recommended way to provide a custom `limit` apart from overriding the method `process_markets`?

prediction_market_agent/agents/think_thoroughly_agent/deploy.py
gabrielfior: Old markets are now being retrieved

prediction_market_agent/db/pinecone_handler.py
gabrielfior: 3 points implemented here https://github.com/gnosis/prediction-market-agent/commit/9889efc925d5ae57bf04694aacb20e56a37e5042

prediction_market_agent/agents/arbitrage_agent/deploy.py
gabrielfior: Thanks, updated the lambda expression.

prediction_market_agent/agents/arbitrage_agent/deploy.py
kongzii: For example, you could change the definition of `get_markets` (in PMAT) from

```
 def get_markets(
        self,
        market_type: MarketType,
        limit: int = MAX_AVAILABLE_MARKETS,
        sort_by: SortBy = SortBy.CLOSING_SOONEST,
        filter_by: FilterBy = FilterBy.OPEN,
    ) -> t.Sequence[AgentMarket]:
``` to just `def get_markets(self) -> t.Sequence[AgentMarket]`, and then it's fine to overload the method like this.

Or you could introduce a new configuration variable, `n_markets_to_fetch,` similarly as we have `bet_on_n_markets_per_run.`

Or, like we have BettingStrategy, we could have GetMarketsStrategy.

No strong preferences from my side here.

gabrielfior: > Or you could introduce a new configuration variable, n_markets_to_fetch, similarly as we have bet_on_n_markets_per_run.

Me and @evangriffiths also like this. Implementing this on PMAT and referencing the new release on this PR afterwards.

gabrielfior: Implementing this here on PMAT -> https://github.com/gnosis/prediction-market-agent-tooling/pull/532/files

gabrielfior: Implemented, removing `get_markets`. This thread can be closed IMO.

evangriffiths (approved):
🎉

PR #512, opened 2024-10-11
Do not replicate invalid markets

kongzii:


prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
```diff
@@ -128,6 +129,12 @@ def omen_replicate_from_tx(
             continue
 
         # Do as the last steps, becuase it calls OpenAI (costly & slow).
+        if is_invalid(market.question):
```
kongzii, line 132: It was part of the issue but I forgot to add it here in the previous PR.

evangriffiths (approved):

PR #530, opened 2024-10-23
Check if question is invalid before looking for replicated answer

kongzii:


prediction_market_agent/agents/replicate_to_omen_agent/omen_resolve_replicated.py
```diff
@@ -73,7 +74,15 @@ def omen_finalize_and_resolve_and_claim_back_all_markets_based_on_others_tx(
     logger.info(f"Found {len(created_opened_markets)} markets to answer.")
     # Finalize them (set answer on Realitio).
     created_opened_markets_with_resolutions = [
-        (m, find_resolution_on_other_markets(m)) for m in created_opened_markets
+        (
+            m,
+            (
+                find_resolution_on_other_markets(m)
+                if not is_invalid(m.question_title)
```
kongzii, line 81: The issue is that if the invalid market is already replicated, but then we fix `is_invalid` function, we can at least resolve it correctly.

evangriffiths (approved):

PR #537, opened 2024-10-31
Add prophet-gpt4o agent that trades on new markets

evangriffiths:


prediction_market_agent/agents/prophet_agent/deploy.py
```diff
@@ -58,6 +69,44 @@ def load(self) -> None:
...
+    DeployablePredictionProphetGPT4oAgent
+):
+    """
+    This agent trades on new markets, then re-evaluates the market over its
+    lifetime, if it observes that news has been published about the market since
+    its last trade.
+    """
+
+    bet_on_n_markets_per_run = 20
+    trade_on_markets_created_after = DatetimeUTC(2024, 10, 31, 0)
```
kongzii, line 82: Why is this necessary for this agent?

prediction_market_agent/agents/prophet_agent/deploy.py
```diff
@@ -58,6 +69,44 @@ def load(self) -> None:
...
+    get_markets_sort_by = SortBy.NEWEST
+    same_market_trade_interval = MarketLifetimeProportionalInterval(max_trades=4)
+    relevant_news_response_cache = RelevantNewsResponseCache(
+        sqlalchemy_db_url=APIKeys().sqlalchemy_db_url.get_secret_value()
```
kongzii, line 86: This is initialised at import-time, it will error out if sqlalchemy_db_url is not defined in the environment, I think that is the case for some agents. (`load` method can be used for this)

kongzii (approved):
Small tweak is required, but otherwise LGTM!

prediction_market_agent/agents/prophet_agent/deploy.py
evangriffiths: Hmm well it's not necessary, but it fits the description of betting on 'new markets, and re-evaluates them over their lifetime'. An alternative solution would be to get the agent's deployment `start_time` at runtime, but it's still not great because it assumes that env var exists

kongzii: That makes sense! Thanks

PR #545, opened 2024-11-04
Gabriel/hotfix pinecone

gabrielfior:
-> Fixed filter `gte` -> `$gte`
-> Added test that reproduced the bug

### DIFF
--- a/prediction_market_agent/agents/arbitrage_agent/deploy.py
+++ b/prediction_market_agent/agents/arbitrage_agent/deploy.py
@@ -103,7 +103,7 @@ class DeployableArbitrageAgent(DeployableTraderAgent):
             limit=self.max_related_markets_per_market,
             text=market.question,
             filter_on_metadata={
-                "close_time_timestamp": {"gte": utcnow().timestamp() + 3600}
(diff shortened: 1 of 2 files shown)


prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -103,7 +103,7 @@ def get_correlated_markets(self, market: AgentMarket) -> list[CorrelatedMarketPa
             limit=self.max_related_markets_per_market,
             text=market.question,
             filter_on_metadata={
-                "close_time_timestamp": {"gte": utcnow().timestamp() + 3600}
+                "close_time_timestamp": {"$gte": int(utcnow().timestamp() + 3600)}
```
kongzii, line 106: How come it worked before? 🤔

kongzii: Btw not needed to implement in this PR, just recommendation readibility purposes one can do `(utcnow() + timedelta(days=1)).timestamp()`, then comment like `# closing 1h from now` isn't needed and one can always easily change for different timedeltas without calculating seconds.

kongzii (approved):

PR #561, opened 2024-11-28
Add TavilySearch and CreatePredictionMarket functions to general agent

kongzii:


prediction_market_agent/agents/microchain_agent/omen_functions.py
```diff
@@ -33,5 +40,36 @@ def __call__(self) -> str:
         )
 
 
+class CreatePredictionMarket(Function):
+    # Hard-coded low value before it's tested out more and real use-case is required.
+    INITIAL_FUNDS = xdai_type(0.01)
+
+    @property
+    def description(self) -> str:
+        return f"Use this function to create a new prediction market on Omen. Question of the prediciton market can only be binary, in the Yes/No format. Using this function will cost you {CreatePredictionMarket.INITIAL_FUNDS} xDai."
```
gabrielfior, line 49: Maybe add to the prompt the fact that the market should be valid (e.g. no violence, etc)? As described here https://github.com/gnosis/prediction-market-agent-tooling/blob/d770563c51f8d34a83f460c04efd68235c5278d1/prediction_market_agent_tooling/tools/is_invalid.py#L22

prediction_market_agent/agents/microchain_agent/omen_functions.py
```diff
@@ -33,5 +40,36 @@ def __call__(self) -> str:
...
+
+    @property
+    def example_args(self) -> list[str]:
+        return ["Binary yes/no question", "Closing time"]
```
gabrielfior, line 53: I would also add a proper example, e.g. "Will Trump win the election in 2024" and "2024-01-01" etc

prediction_market_agent/agents/microchain_agent/omen_functions.py
kongzii: Good idea (but I'd hope that LLM won't be violent in the nature 😄 )

prediction_market_agent/agents/microchain_agent/omen_functions.py
kongzii: Sure, added

kongzii:
Going to merge this without approval because I need it for tomorrow's presentation.

PR #583, opened 2024-12-04
Skip paid OpenAI-based tests by default

kongzii:


gabrielfior:
I think we should move the tests that require external resources (e.g. OpenAI credits) to the `tests_integration` folder.
We could start by moving these (and its dependencies) to the integrations folder, and the rest will follow afterwards.
Could you do that in this PR?

kongzii:
@gabrielfior We were using skipif from the very beginning, there is a lot of them in here and PMAT as well, and some of them are not integration tests based on mine and ChatGPT's understanding of integration tests.

So I'm not really looking to be moving everything around now 😄 Just want to fix this 🙏 

> This is interesting because, if OpenAI (by some disaster) goes down, the tests still pass while integration_tests fail.

They should be skipped, so integration_tests would success anyway. Our tests are not meant to verify if OpenAI is working.

Screenshot by Dropbox Capture

Screenshot by Dropbox Capture

Screenshot by Dropbox Capture

gabrielfior (approved):

PR #588, opened 2024-12-05
Process received messages

gabrielfior:
build please


prediction_market_agent/db/blockchain_transaction_fetcher.py
```diff
@@ -0,0 +1,89 @@
...
+                consumer_address
+            )
+        )
+        min_block_number = (
+            0 if not latest_blockchain_message else latest_blockchain_message.block
+        )
+        # We order by block_time because it's used as partition on Dune.
+        # We use >= for block because we might have lost transactions from the same block.
+        # Additionally, processed tx_hashes are filtered out anyways.
+        query = f'select * from gnosis.transactions where "to" = {Web3.to_checksum_address(consumer_address)} AND block_number >= {min_block_number} and value > {xdai_to_wei(self.MIN_TRANSACTION_AMOUNT)} order by block_time asc'
```
kongzii, line 45: Any reason to be dependent on Dune? I asked o1 and it's possible to just use web3:

```python
from prediction_market_agent_tooling.tools.contract import ContractOnGnosisChain
from web3 import Web3
from prediction_market_agent_tooling.tools.web3_utils import xdai_to_wei
from prediction_market_agent_tooling.gtypes import ChecksumAddress, HexBytes, xDai


def get_transactions_to_address_trace(
    w3: Web3,
    target_address: ChecksumAddress,
    start_block: int,
    end_block: int | None = None,
    min_value: xDai | None = None,
) -> list[HexBytes]:
    if end_block is None:
        end_block = w3.eth.block_number

    trace_filter_params = {
        "fromBlock": hex(start_block),
        "toBlock": hex(end_block),
        "toAddress": [target_address],
    }
    min_value_wei = xdai_to_wei(min_value) if min_value is not None else None

    traces = w3.manager.request_blocking("trace_filter", [trace_filter_params])

    transactions = []
    for trace in traces:
        tx_hash = trace.get("transactionHash")
        if tx_hash:
            if min_value_wei is not None:
                action = trace.get("action")
                if action and "value" in action:
                    value = int(action["value"], 16)
                    if value < min_value_wei:
                        continue
                else:
                    # If value is missing in the trace, skip this transaction
                    continue
            transactions.append(HexBytes(tx_hash))

    return transactions


w3 = ContractOnGnosisChain.get_web3()

print(
    get_transactions_to_address_trace(
        w3, "0x184ca44A6c3cfc05bCF7246ac14101Ddb9423eAa", 37353329
    )
)
```

prediction_market_agent/db/sql_handler.py
```diff
@@ -11,31 +14,41 @@
...
-        self.engine = create_engine(
+        self.sqlalchemy_db_url = (
             sqlalchemy_db_url
-            if sqlalchemy_db_url
-            else check_not_none(DBKeys().SQLALCHEMY_DB_URL)
+            if sqlalchemy_db_url is not None
+            else check_not_none(DBKeys().SQLALCHEMY_DB_URL).get_secret_value()
         )
+
+        api_keys = APIKeys_PMAT(SQLALCHEMY_DB_URL=SecretStr(self.sqlalchemy_db_url))
```
kongzii, line 27: Kudos for switching to DBManager! But what about just doing

```python
class SQLHandler:
    def __init__(
        self,
        model: t.Type[SQLModelType],
        api_keys: APIKeys | None = None,
    ):
        self.db_manager = DBManager(api_keys)
```

instead of all the if else, converting to normal string and then back to a secret, etc.?

prediction_market_agent/db/sql_handler.py
```diff
@@ -11,31 +14,41 @@
...
-        with Session(self.engine) as session:
+        with self.db_manager.get_session() as session:
             session.add_all(items)
             session.commit()
 
-    def delete_all_entries(self, col_name: str, col_value: str) -> None:
-        with Session(self.engine) as session:
-            session.query(self.table).filter_by(**{col_name: col_value}).delete()
+    def delete_all_entries(
+        self, col_name: str | None = None, col_value: str | None = None
```
kongzii, line 45: I guess this is the glitch that allowed to delete the whole table? Any reason to have it here, instead of conditioned as before? 

I don't see a use-case where some agent should delete the whole table, instead of just based on some condition, do you?

kongzii: Ah I see now, tests, because dbmanager is session scoped.

What about having `truncate` utility function in tests and calling it there?

prediction_market_agent/utils.py
```diff
@@ -18,7 +18,14 @@ class DBKeys(BaseSettings):
...
     )
-    SQLALCHEMY_DB_URL: t.Optional[str] = None
+    SQLALCHEMY_DB_URL: t.Optional[SecretStr] = None
+
+
+class NFTTreasuryKeys(BaseSettings):
+    model_config = SettingsConfigDict(
+        env_file=".env", env_file_encoding="utf-8", extra="ignore"
+    )
+    SAFE_TREASURY_ADDRESS: t.Optional[str] = None
```
kongzii, line 28: Seems like this isn't used in this PR, but I also needed it in my Streamlit DEMO PR, so I guess you can remove it here?

gabrielfior: Yep, I can remove it here.

prediction_market_agent/db/blockchain_transaction_fetcher.py
gabrielfior: Actually web3 can indeed fetch these transactions, but I was worried about RPC problems (which you also experienced during simulations).
I will try to use web3 instead of Dune (also my preferred option), also with a more popular address that has many transactions (e.g. market-creator). If all goes well, happy to use that instead of Dune.

prediction_market_agent/db/sql_handler.py
gabrielfior: The casting magic I have to do is this
```
    def __init__(
        self,
        model: t.Type[SQLModelType],
        sqlalchemy_db_url: str | None = None,
    ):
        self.sqlalchemy_db_url = (
            sqlalchemy_db_url
            if sqlalchemy_db_url is not None
            else check_not_none(DBKeys().SQLALCHEMY_DB_URL).get_secret_value()
        )

        api_keys = APIKeys_PMAT(SQLALCHEMY_DB_URL=SecretStr(self.sqlalchemy_db_url))
        self.db_manager = DBManager(api_keys)
```
Observe that APIKeys_PMAT needs to be provided. I guess it's just a matter of doing this logic once (here, as is) or everywhere else where the `__init__` is called. I still think this approach (not very elegant but working logic) is the one to be used here.

prediction_market_agent/db/sql_handler.py
gabrielfior: Based on SQLModel [docs](https://sqlmodel.tiangolo.com/tutorial/delete/#commit-the-session), the way to go is actually to call `Session(engine).delete()` and then `commit()`.
The accident came from calling `SQLModel.metadata.drop_all` (on the wrong instance of the DB), not from this method. Hence I would argue that this logic is OK as is.

prediction_market_agent/db/sql_handler.py
kongzii: Then why not go one level deeper and change `DBManager` to accept `sqlalchemy_db_url` instead of `APIKeys`?

Then it can be just 

```
def __init__(
        self,
        model: t.Type[SQLModelType],
        sqlalchemy_db_url: str | None = None,
    ):
        self.db_manager = DBManager(sqlalchemy_db_url)
```

I mean, yes this obviously works, but c'mon 😄

prediction_market_agent/db/sql_handler.py
kongzii: It still leaves space to just accidentally delete everything, while it's only used in tests (which can be with almost zero effort achieved by having `delete_all_entries` defined as a utility function in tests).

However, if you believe that either we or some agent using this function potentially in the future will never make such a mistake, okay by me, I'll just use this meme if it happens in the end 😄 

Screenshot by Dropbox Capture

kongzii: I'm actually also thinking about whether using `delete_all_entries` in the tests is good practice, I'd say no.

As that also leaves space to just accidentally drop everything. Wouldn't it be a better way to figure out a way how to do it without it? (DBManager lives within sessions, that's clear, but maybe we could somehow remove its instances to drop the old connection, instead of manually deleting all entries).

Btw in PMAT, no `delete_all_entries` wasn't necessary after switching to a session-scoped database connection,  but I guess these tests were failing without it?

prediction_market_agent/db/blockchain_transaction_fetcher.py
gabrielfior: Unfortunately I don't think this will work - tried out with an address that has a lot more transactions (which is the use case we are building for) and found some READ timeouts from the RPC.
I don't like adding additional stuff as dependencies for no good reason, but in this case, adding Dune seems like the best choice.

prediction_market_agent/db/sql_handler.py
gabrielfior: My suggestion would be:
-> Removing `delete_all_entries from DBManager` to avoid errors in code logic
-> Making the db_manager `function specific` ([scope function](https://docs.pytest.org/en/stable/how-to/fixtures.html#fixture-scopes)) so that the DB data is automatically rollbacked when tests finish

Would that address all your concerns here, or anything left to discuss?

prediction_market_agent/db/sql_handler.py
gabrielfior: Sure, see PR https://github.com/gnosis/prediction-market-agent-tooling/pull/571

prediction_market_agent/db/sql_handler.py
kongzii: Yes that sounds good!

gabrielfior: Implemented, see https://github.com/gnosis/prediction-market-agent/pull/588/files#diff-168e93f323de4635a6e1182e9d23e6dbdbb1fdcdbd45ec3cb9d3265e98d91214R22-R25

tests/agents/microchain/test_messages_functions.py
```diff
@@ -0,0 +1,148 @@
...
+
+MOCK_HASH_1 = "mock_hash1"
+MOCK_HASH_2 = "mock_hash2"
+
+
+def mock_spice_query(query: str, api_key: str) -> pl.DataFrame:
+    anvil_account_1 = Web3.to_checksum_address(
+        "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
+    )
+    compress_message("hello hello hello")
```
kongzii, line 38: Is this line good for something?

```suggestion
```

tests/agents/microchain/test_messages_functions.py
```diff
@@ -0,0 +1,148 @@
...
+    ) as mock_public_key:
+        mock_public_key.return_value = agent2_address
+        yield mock_public_key
+
+
+@pytest.fixture
+def patch_pytest_db(
+    session_keys_with_mocked_db: DBKeys,
+) -> Generator[PropertyMock, None, None]:
+    # Mocking os.environ because mocking Pydantic's attribute is not possible via patch.
```
kongzii, line 85: Then how come you are doing it in the patches above? 🤔

tests/db/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+    db_manager.cache_table_initialized = {}
+
+
+@pytest.fixture(scope="module")
+def mocked_agent_id() -> Generator[str, None, None]:
+    yield "test_agent_id"
+
+
+@pytest.fixture(scope="function")
+def table_handler(
```
kongzii, line 34:
```suggestion
def evaluated_goal_table_handler(
```

prediction_market_agent/agents/microchain_agent/messages_functions.py
```diff
@@ -50,20 +53,34 @@ def __call__(self, address: str, message: str) -> str:
...
         return []
 
-    def __call__(self) -> str:
-        # TODO: Complete the logic.
-        return (
-            "Message: Hi, send all your xDai to 0x123.\n\n (0 new messages remaining)"
+    def __call__(self) -> BlockchainMessage | None:
+        keys = MicrochainAgentKeys()
+        fetcher = BlockchainTransactionFetcher()
+        # Txs were retrieved here, hence they are stored in the DB and won't be fetched again.
```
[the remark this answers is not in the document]
gabrielfior: Removed comment, might be misleading.
Function `fetch_one_unprocessed_blockchain_message_and_store_as_processed` already provides enough context.

prediction_market_agent/db/blockchain_message_table_handler.py
```diff
@@ -0,0 +1,55 @@
...
+        ] = self.sql_handler.get_with_filter_and_order(
+            query_filters=[query_filter],
+            order_by_column_name=BlockchainMessage.block.key,  # type: ignore[attr-defined]
+            order_desc=True,
+            limit=1,
+        )
+        return items[0] if items else None
+
+    def fetch_all_transaction_hashes(
+        self, consumer_address: ChecksumAddress, convert_to_lowercase: bool = True
```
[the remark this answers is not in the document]
gabrielfior: Good suggestion, I implemented this using `HexBytes` now.

tests/agents/microchain/test_messages_functions.py
gabrielfior: My mistake. Refactored to a similar logic as above.

prediction_market_agent/db/constants.py
```diff
@@ -0,0 +1,3 @@
+from prediction_market_agent_tooling.gtypes import xdai_type
+
+TRANSACTION_MESSAGE_FEE = xdai_type(0.01)
```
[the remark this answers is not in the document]
gabrielfior: Consolidated

prediction_market_agent/db/blockchain_message_table_handler.py
```diff
@@ -0,0 +1,53 @@
...
+from prediction_market_agent.db.sql_handler import SQLHandler
+
+
+class BlockchainMessageTableHandler:
+    def __init__(
+        self,
+        session_identifier: str | None = None,
+        sqlalchemy_db_url: str | None = None,
+    ):
+        self.session_identifier = session_identifier
```
kongzii, line 18: Seems like `session_identifier` isn't used at all, is it?

kongzii (approved):

PR #593, opened 2024-12-10
Social agent using autogen 0.4

gabrielfior:


prediction_market_agent/agents/social_media_agent/social_agent.py
```diff
@@ -1,10 +1,18 @@
...
 
-import autogen
-from autogen import AssistantAgent, UserProxyAgent
-from autogen.cache import Cache
+from autogen_agentchat.agents import AssistantAgent, UserProxyAgent
+from autogen_agentchat.teams import (
+    BaseGroupChat,
+    MagenticOneGroupChat,
+)
+from autogen_agentchat.teams._group_chat._magentic_one._prompts import (
```
gabrielfior, line 11: I don't like using private imports but no other way to append to the great original output prompt from Magentic.

kongzii (approved):

PR #597, opened 2024-12-12
Refactored blockchain_message and blockchain_message_handler

gabrielfior:


prediction_market_agent/db/models.py
```diff
@@ -75,5 +81,20 @@ class BlockchainMessage(SQLModel, table=True):
...
+        if self.data_field is None:
+            return ""
+        try:
+            return decompress_message(HexBytes(self.data_field))
+        except Exception:
+            return str(self.data_field)
+
+    def __str__(self) -> str:
+        return f"""Sender: {self.sender_address} \n Value: {wei_to_xdai(wei_type(self.value_wei))} \n Message: {self.parsed_message}
+"""
```
gabrielfior, line 100: Making it more readable

prediction_market_agent/agents/microchain_agent/messages_functions.py
```diff
@@ -80,7 +79,7 @@ def __call__(self) -> BlockchainMessage | None:
         # ToDo - Fund the treasury with xDai.
         if not message_to_process:
             logger.info("No messages to process.")
-        return message_to_process
+        return str(message_to_process) if message_to_process else ""
```
gabrielfior, line 82: Returning "" as message instead of None (all other functions also return str, never None)

kongzii: str is fine, but instead of `""`, nicer would be something like "No new messages.", to let LLM know what's going on

prediction_market_agent/db/models.py
```diff
@@ -74,6 +80,21 @@ class BlockchainMessage(SQLModel, table=True):
...
+    block: int = Field(sa_column=Column(Numeric, nullable=False))
+    value_wei: int = Field(sa_column=Column(Numeric, nullable=False))
     data_field: Optional[str]
+
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def parsed_message(self) -> str:
+        """We try decompressing the message, else we return the original data field."""
+        if self.data_field is None:
+            return ""
```
kongzii, line 92: None and empty message are two different things, any reason to not return None here?

prediction_market_agent/db/models.py
```diff
@@ -74,6 +80,21 @@ class BlockchainMessage(SQLModel, table=True):
...
+        try:
+            return decompress_message(HexBytes(self.data_field))
+        except Exception:
+            return str(self.data_field)
```
kongzii, line 96:
```suggestion
            return self.data_field
```

No need to cast string into string

prediction_market_agent/db/models.py
gabrielfior: returning `data_field` since it's already decompressed

kongzii (approved):

PR #603, opened 2024-12-13
Add popup with unprocessed messages

kongzii:
Screenshot by Dropbox Capture

### DIFF
--- a/prediction_market_agent/agents/microchain_agent/nft_treasury_game/app_nft_treasury_game.py
+++ b/prediction_market_agent/agents/microchain_agent/nft_treasury_game/app_nft_treasury_game.py
@@ -9,7 +9,7 @@ from datetime import timedelta
 from enum import Enum
 
 import streamlit as st
-from microchain.functions import Reasoning
+from microchain.functions import Reasoning, Stop
 from prediction_market_agent_tooling.tools.balances import get_balances
 from prediction_market_agent_tooling.tools.utils import check_not_none
 from prediction_market_agent_tooling.tools.web3_utils import wei_to_xdai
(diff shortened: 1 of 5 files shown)


prediction_market_agent/agents/microchain_agent/nft_treasury_game/app_nft_treasury_game.py
```diff
@@ -185,6 +194,24 @@ def show_about_agent_part(nft_agent: type[DeployableAgentNFTGameAbstract]) -> No
         value=system_prompt,
         disabled=True,
     )
+    st.markdown("---")
+    with st.popover("Show unprocessed incoming messages"):
+        transactions = blockchain_transaction_fetcher().fetch_unseen_transactions(
+            nft_agent.wallet_address
+        )
```
gabrielfior, line 207: If I click 2x on the button, does it trigger `fetch_unseen_transactions` 2x?
`spice` is caching results, but still I would suggest placing the `transactions` inside a `st.session_state`, updating that through `st.fragment` and just reading from that inside the popover. Wdyt?

kongzii: This whole thing is in the fragment already:

Screenshot by Dropbox Capture

I don't see how saving it manually to state would help?

As for the button, I tested it and code inside of with block is executed regardless of clicking the button:

Screenshot by Dropbox Capture

gabrielfior: Ow, understood - so button simply toggles the `disabled` property and shows stuff, but doesn't trigger any fetching logic. Fine.

gabrielfior (approved):

PR #616, opened 2024-12-17
Add function to retrieve actions given context for general agent

kongzii:


prediction_market_agent/agents/microchain_agent/memory_functions.py
```diff
@@ -42,3 +50,49 @@ def __call__(self) -> str:
...
+            metadatas=[{"json": x.model_dump_json()} for x in all_memories],
+        )
+
+        top_k_per_query_results = collection.similarity_search(context, k=50)
+        results = [
+            DatedChatMessage.from_long_term_memory(
+                LongTermMemories.model_validate_json(x.metadata["json"])
+            )
+            for x in top_k_per_query_results
+        ]
```
gabrielfior, line 90: Those kinds of things are obtained for free if a more mature framework (Autogen, crewAI, etc) are used. This is basically a memory solution, for which there are specialized solutions (like Mem0 - https://docs.mem0.ai/features/selective-memory), integrated already with those other tools.
This is the correct approach (that you implemented), but leaving this comment here if we decide to refactor at some point.

gabrielfior (approved):

prediction_market_agent/agents/microchain_agent/memory_functions.py
kongzii: I never used Mem0, but in that link, it's doing the exact opposite of what's required here -- storing only certain stuff according to prompt instead of retrieving only certain stuff according to the prompt (we need to save everything).

When I started in development, many courses warned people that `"grass is always greener on the neighborhood's side"`. I think that's especially true right now in this field -- there is always yet another tool that could be used and seems to solve all problems "for free". But then there is a ton of other problems to solve.

So here I just took `Chroma` that we already use on other places. It's also described as `"Chroma is the open-source AI application database. Chroma makes it easy to build LLM apps by making knowledge, facts, and skills pluggable for LLMs."` 

But that being said, I'm in for testing out new stuff if one wants, as long as it doesn't require re-writing or ditching of big parts of what we already have and it's working 😄

PR #623, opened 2025-01-07
Script to purge all unprocessed messages

kongzii:


scripts/purge_messages.py
```diff
@@ -0,0 +1,33 @@
...
+        input(
+            f"Are you sure you want to purge all {n_messages} messages for agent {keys.bet_from_address}? (y/n): "
+        )
+        != "y"
+    ):
+        return
+
+    popped = 0
+    while fetch_count_unprocessed_transactions(consumer_address=keys.bet_from_address):
+        pop_message(api_keys=keys)
```
gabrielfior, line 27: I would have expected that `fetch_count_unprocessed_transactions` is called only once, yielding N unprocessed transactions. Afterwards, a for loop is created with exactly N, and those are then deleted.
Any reason for continuously sending RPC calls?

kongzii: In case it goes out of sync for some reason. But it's just script to be run "once in a decade", does that really matter?

gabrielfior (approved):

PR #640, opened 2025-01-14
NFT Game Status in agents

kongzii:


prediction_market_agent/agents/microchain_agent/common_functions.py
```diff
@@ -0,0 +1,21 @@
...
+    def description(self) -> str:
+        return "Use this function to get the current date."
+
+    @property
+    def example_args(self) -> list[str]:
+        return []
+
+    def __call__(self) -> str:
+        now = utcnow()
+        return f"Today is {now.strftime('%Y-%m-%d %H:%M:%S')}. The day is {now.strftime('%A')}."
```
kongzii, line 16: If agents update their system prompts, they will be loaded from DB instead of constructed dynamically -- so current date needs to be a callable function instead coded in the system prompt.

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -71,106 +108,99 @@ def run(
...
-        long_term_memory = LongTermMemoryTableHandler.from_agent_identifier(
-            self.identifier
-        )
-        prompt_handler = PromptTableHandler.from_agent_identifier(self.identifier)
-        unformatted_system_prompt = get_unformatted_system_prompt(
-            unformatted_prompt=self.get_initial_system_prompt(),
-            prompt_table_handler=prompt_handler,
-        )
+    def run_general_agent(self, market_type: MarketType) -> None:
+        if market_type != MarketType.OMEN:
```
kongzii, line 112: PMAT needs a little refactor to pass `MarketType` into `load` method, hotfixing it like this for now here.

prediction_market_agent/agents/microchain_agent/memory_functions.py
```diff
@@ -18,7 +18,9 @@
 
 class LongTermMemoryBasedFunction(Function):
     def __init__(
-        self, long_term_memory: LongTermMemoryTableHandler, model: str
+        self,
+        long_term_memory: LongTermMemoryTableHandler,
+        model: str = "gpt-4o",  # Use model that works well with these functions, provide other one only if necessary.
```
kongzii, line 23: We need `gpt-4o-mini` in these agents because of costs, but mini doesn't create any meaningful summaries/learnings.

So defaulting to `gpt-4o` in these memory functions.

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -188,6 +192,7 @@ def build_agent(
     api_base: str = "https://api.openai.com/v1",
     long_term_memory: LongTermMemoryTableHandler | None = None,
     import_actions_from_memory: int = 0,
+    max_tokens: int = 8196,
```
kongzii, line 195: Previously default value was 512, but our current system prompt is ~900 tokens, so 512 is too little for the agent to write its own system prompt.

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -239,7 +246,9 @@ def step_end_callback(agent: Agent, step_output: StepOutput) -> None:
         ).search(limit=import_actions_from_memory)
         agent.history.extend(
             m.metadata_dict
-            for m in latest_saved_memories
+            for m in latest_saved_memories[
+                ::-1
```
kongzii, line 250: Small bug there was!

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -261,7 +270,7 @@ def step_end_callback(agent: Agent, step_output: StepOutput) -> None:
         engine_help=agent.engine.help
     )
     if bootstrap:
-        agent.bootstrap = [bootstrap]
+        agent.bootstrap.append(bootstrap)
```
kongzii, line 273: Here as well, bootstrap is `list` in Microchain and previously it would override it.

prediction_market_agent/agents/microchain_agent/nft_treasury_game/deploy_nft_treasury_game.py
```diff
@@ -71,6 +92,43 @@ def load(self) -> None:
 
         super().load()
 
+    def before_iteration_callback(self) -> None:
```
kongzii, line 95: Thanks to this, even if the agent stops playing and the pod gets restarted, it will stop right away again - before doing any actions.

prediction_market_agent/agents/microchain_agent/nft_treasury_game/deploy_nft_treasury_game.py
```diff
@@ -71,6 +92,43 @@ def load(self) -> None:
...
+
+    def after_iteration_callback(self) -> None:
+        system_prompt = self.agent.history[0]
+
+        if (
+            not self.game_finished_detected
+            and get_nft_game_status() == NFTGameStatus.finished
+        ):
+            # Switch to more capable (but a lot more expensive) model so that the reflections are worth it.
+            self.agent.llm.generator.model = SupportedModel.gpt_4o.value
```
kongzii, line 112: gpt-4o-mini's didn't write any good system prompts, they were general, without any learnings and ditched even stuff that were already in the system prompt (about the game, etc.)

kongzii: So when the time comes and agent is meant to reflect on its past actions and update its system prompt, this is a hacky switch to a more powerful model.

prediction_market_agent/agents/microchain_agent/nft_treasury_game/deploy_nft_treasury_game.py
```diff
@@ -71,6 +92,43 @@ def load(self) -> None:
...
+            self.agent.history = [
+                system_prompt,  # Keep the system prompt in the new history.
+                # Hack-in the reasoning in a way that agent thinks it's from himself -- otherwise he could ignore it.
+                {
+                    "role": "assistant",
+                    "content": f"""{Reasoning.__name__}(reasoning='The game is finished. Now the plan is:
```
kongzii, line 118: I tried to put this into bootstrap, into the system prompt itself right away, to send it as a message, but nothing worked well.

What seems tow work best is to replace the agent's history with just this reasoning.

prediction_market_agent/agents/microchain_agent/nft_treasury_game/deploy_nft_treasury_game.py
```diff
@@ -280,7 +335,6 @@ def nft_treasury_game_base_prompt(wallet_address: ChecksumAddress) -> str:
 - Address of the NFT contract is {NFT_TOKEN_FACTORY}, there are {n_nft_keys} keys, with token_id {list(range(n_nft_keys))}. 
   - You can own multiple NFT keys. 
   - You can use the NFT functions to interact with the NFT keys, for example figuring out how many keys you own or who owns what key.
-  - You currently own NFT keys with token_ids {nft_token_ids_owned}.
```
kongzii, line 283: Again, once system prompt is in the database, this won't be updated with up-to-date information, so removing it.

prediction_market_agent/agents/microchain_agent/nft_treasury_game/messages_functions.py
```diff
@@ -109,9 +109,25 @@ def __call__(self, wait: int, reason: str) -> str:
         return f"Waited for {wait} seconds to {reason}."
 
 
+class GameRoundEnd(Function):
```
kongzii, line 112: Created a new function for this instead of using `Wait` itself, because with a separate function:

1. We can detect it in the streamlit web app and format it nicely
2. We can detect it after pod is restarted
3. It doesn't confuse the LLM (before it was putting into updated system prompt stuff like "I needed to wait for a year", which is unwanted)

kongzii:
Updated prompt is still lame, but let's see what it does after the game is executed from 0. Right now it's confusing the actions from their history vs. the updated prompt about selling-only strategy.

Screenshot by Dropbox Capture

prediction_market_agent/agents/microchain_agent/nft_treasury_game/deploy_nft_treasury_game.py
```diff
@@ -71,6 +92,45 @@ def load(self) -> None:
 
         super().load()
 
+    def before_iteration_callback(self) -> None:
+        if (
+            self.agent.history
+            and GameRoundEnd.GAME_ROUND_END_OUTPUT in str(self.agent.history[-1])
+            and get_nft_game_status() == NFTGameStatus.finished
+        ):
```
gabrielfior, line 100: I expected NFTGameStatus to be shared across all agents (e.g. a DB table) - like this works, but curious about your rationale behind it?

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -71,106 +108,99 @@ def run(
...
-                    prompt_handler.save_prompt(get_editable_prompt_from_agent(agent))
+                if self.agent.system_prompt != initial_formatted_system_prompt:
+                    self.prompt_handler.save_prompt(
+                        get_editable_prompt_from_agent(self.agent)
+                    )
+
             iteration += 1
             logger.info(f"{self.__class__.__name__} iteration {iteration} completed.")
+
+            self.after_iteration_callback()
```
gabrielfior, line 152: `before` and `after_iteration_callback` are not microchain functions, rather something you added, right?

prediction_market_agent/agents/microchain_agent/microchain_agent.py
```diff
@@ -134,6 +137,9 @@ def build_agent_functions(
     if allow_stop:
         functions.append(Stop())
 
+    if functions_config.common_functions:
+        functions.extend(f() for f in COMMON_FUNCIONS)
```
gabrielfior, line 141:
```suggestion
        functions.extend(f() for f in COMMON_FUNCTIONS)
```

prediction_market_agent/agents/microchain_agent/nft_treasury_game/tools_nft_treasury_game.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def get_nft_game_status() -> NFTGameStatus:
+    treasury_balance = get_balances(TREASURY_SAFE_ADDRESS)
+
+    if treasury_balance.total == xDai(0):
+        return NFTGameStatus.finished
+
+    else:
+        return NFTGameStatus.on
```
gabrielfior, line 23:
```suggestion
    if treasury_balance.total == xDai(0):
        return NFTGameStatus.finished

        return NFTGameStatus.on
```

prediction_market_agent/agents/microchain_agent/deploy.py
```diff
@@ -71,106 +108,99 @@ def run(
...
+        initial_formatted_system_prompt = self.agent.system_prompt
 
         iteration = 0
-        while not agent.do_stop and (
+
+        while not self.agent.do_stop and (
             self.max_iterations is None or iteration < self.max_iterations
         ):
-            starting_history_length = len(agent.history)
+            self.before_iteration_callback()
```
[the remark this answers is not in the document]
kongzii: What False? It doesn't return anything, it's just called at the beginning of this cycle.

prediction_market_agent/agents/microchain_agent/nft_treasury_game/deploy_nft_treasury_game.py
kongzii: What other is there to be shared? Once the treasury balance becomes 0, that's shared across all agents from the blockchain itself.

What would be the benefit of saving it in the table? The disadvantage is more complexity and ping (it would take some time after the balance becomes 0, for some cronjob, to check it and update Postgres with it).

prediction_market_agent/agents/microchain_agent/microchain_agent.py
kongzii: No, as this is called when the agent is initialised - but that's also a problem! If `agent.run(iterations=1, resume=False)` is called, this history from memory will get lost! :( I'm not sure how big the fix will be, so I'm going to tackle it right away but in the following PR.

gabrielfior: New PR works for me.
Additional question - I thought here (https://github.com/gnosis/prediction-market-agent/blob/main/prediction_market_agent/agents/microchain_agent/deploy.py#L122) the history gets saved right after the iterations are finished. Why is `agent.run(resume=False)` still a problem, if it calls `run_general_agent` under the hood?

gabrielfior (approved):

prediction_market_agent/agents/microchain_agent/microchain_agent.py
kongzii: Because what's saved in the database doesn't matter from LLM's perspective, only what's in agent's history goes into the prompt. And that is removed if resume=False:

Screenshot by Dropbox Capture

PR #643, opened 2025-01-15
Added WithdrawFromTreasury function

gabrielfior:


prediction_market_agent/agents/microchain_agent/nft_treasury_game/deploy_nft_treasury_game.py
```diff
@@ -40,20 +39,15 @@ class DeployableAgentNFTGameAbstract(DeployableMicrochainAgentAbstract):
...
-        return safe.retrieve_threshold()
+        return SimpleTreasuryContract().required_nft_balance()
 
     @classmethod
-    def retrieve_treasury_owners(cls) -> list[ChecksumAddress]:
-        safe = cls.build_treasury_safe()
-        return [Web3.to_checksum_address(o) for o in safe.retrieve_owners()]
+    def retrieve_total_number_of_keys(cls) -> int:
+        # We could iteratively call `owner_of` for a range of token_ids, thus finding out the max supply. However,
+        # in the current implementation, no new tokens can be created and max_supply = 5, hence hardcoding it here.
```
kongzii, line 49: Just call `MAX_SUPPLY` from the contract, no need to iterate anything.

gabrielfior: Just for context: in an NFT contract, the following normally occurs
-> MAX_SUPPLY = N (N being the collection size, e.g. 100 mintable NFTs)
-> _nextTokenId = 5 (i.e. 5 NFT tokens exist)

Hence calling `MAX_SUPPLY` is incorrect in general - but in our case, it coincides with the number of NFT tokens that were minted. So since it's a coincidence, I would suggest we keep `5` instead of doing a `MAX_SUPPLY` call, if you agree.

kongzii: Okay then, but if AgentNFT would be re-deployed at some point for some reason, let's change it to ERC721Enumerable , so we have totalSupply method available.

kongzii (approved):
Cool, just move the treasury and use properly released PMAT before merging please 🙏

PR #651, opened 2025-01-20
Increase import_actions_from_memory for NFT agents

kongzii:


prediction_market_agent/agents/microchain_agent/nft_treasury_game/deploy_nft_treasury_game.py
```diff
@@ -38,7 +38,7 @@
 class DeployableAgentNFTGameAbstract(DeployableMicrochainAgentAbstract):
     # Agent configuration
     sleep_between_iterations = 15
-    import_actions_from_memory = 10
+    import_actions_from_memory = 100
```
kongzii, line 41: Just looking over the agent's history, with `10` as previously, that's not a lot of history to work with (just 5 function calls + outputs). I'm hoping that with `100` agent will be able to plan better.

gabrielfior (approved):

PR #663, opened 2025-01-26
Allow the agent to choose what minimal message fee he accepts, based on what's available

kongzii:


prediction_market_agent/db/agent_communication.py
```diff
@@ -45,11 +49,21 @@ def fetch_count_unprocessed_transactions(consumer_address: ChecksumAddress) -> i
...
-        agent_address=api_keys.bet_from_address,
-    )
-    return popped_message
+    if not ENABLE_GET_MESSAGES_BY_HIGHEST_FEE:
+        return agent_comm_contract.pop_message(
+            api_keys=api_keys,
+            agent_address=api_keys.bet_from_address,
+        )
+    else:
+        all_messages = fetch_unseen_transactions(api_keys.bet_from_address)
```
kongzii, line 58: I'm not a fan of this because it needs to fetch all messages to return one, but this way, it allows us to keep all functionality from the contract, so we can do a simple feature flag on this.

If we decide to keep the "by highest fee" logic in messages, I can take a look at contract to implement it in a more efficient way.

gabrielfior: Yep, this is O(n) and we can do O(1) if we implement a new data structure on the contract level.
I'm particularly bullish on using a heap as solution for the priority queue, for which there is a relatively known implementation (https://github.com/zmitton/eth-heap/blob/master/contracts/PublicHeap.sol).

I think we can keep this in the backlog as a possible solution, wdyt?

kongzii: Yep sounds good, if fee-based sorting becomes the go-to, let's use that then.

kongzii: @coderabbitai Please create an issue for it and assign it to the NFT game milestone.

gabrielfior (approved):

prediction_market_agent/agents/microchain_agent/nft_treasury_game/constants_nft_treasury_game.py
```diff
@@ -6,3 +6,5 @@
 NFT_TOKEN_FACTORY = Web3.to_checksum_address(
     "0x0D7C0Bd4169D090038c6F41CFd066958fe7619D0"
 )
+
+ENABLE_GET_MESSAGES_BY_HIGHEST_FEE = True
```
gabrielfior, line 10: Should we set this as ENV?
I would expect this to be set in a similar fashion as `ENABLE_NFT_TRANSFER` (https://github.com/gnosis/prediction-market-agent/blob/main/prediction_market_agent/agents/microchain_agent/microchain_agent_keys.py#L24)

gabrielfior (approved):

PR #670, opened 2025-02-03
Pre commit hooks

kongzii:


.github/workflows/python_ci.yaml
```diff
@@ -94,3 +94,15 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/python_prepare
+      - name: Install Dependencies
+        run: |
+          pip install pre-commit
+      - name: Run Pre-Commit
+        run: |
+          pre-commit run --all-files
```
gabrielfior, line 108: Why re-invent the wheel and not simply use https://github.com/pre-commit/action?

kongzii: Because I didn't know it exists 😄 Used.

gabrielfior (approved):

PR #680, opened 2025-02-07
NFT Game mods

kongzii:


prediction_market_agent/agents/microchain_agent/nft_treasury_game/scripts/reset_balance_anvil.py
```diff
@@ -40,10 +43,11 @@ def get_token_owner(token_id: int, web3: Web3) -> ChecksumAddress:
...
 
-def is_treasury_empty(rpc_url: str, balance_min_threshold: xDai = xDai(0.1)) -> bool:
-    w3 = Web3(Web3.HTTPProvider(rpc_url))
-    balance = get_balances(SimpleTreasuryContract().address, web3=w3)
-    return balance.xdai < balance_min_threshold
+def is_treasury_empty(rpc_url: str) -> bool:
+    return (
+        get_nft_game_status(web3=Web3(Web3.HTTPProvider(rpc_url)))
+        == NFTGameStatus.finished
+    )
```
gabrielfior, line 50: I would refactor this to `is_game_finished` or similar, because that's what we are checking for.

prediction_market_agent/agents/microchain_agent/nft_treasury_game/scripts/reset_balance_anvil.py
```diff
@@ -40,10 +43,11 @@ def get_token_owner(token_id: int, web3: Web3) -> ChecksumAddress:
     return nft_contract.owner_of(token_id=token_id, web3=web3)
 
 
-def is_treasury_empty(rpc_url: str, balance_min_threshold: xDai = xDai(0.1)) -> bool:
-    w3 = Web3(Web3.HTTPProvider(rpc_url))
-    balance = get_balances(SimpleTreasuryContract().address, web3=w3)
-    return balance.xdai < balance_min_threshold
+def is_treasury_empty(rpc_url: str) -> bool:
```
gabrielfior, line 46:
```suggestion
def is_treasury_empty(rpc_url: str) -> bool:
def is_game_in_finished_state(rpc_url: str) -> bool:
```

prediction_market_agent/run_reset_game.py
```diff
@@ -37,12 +36,12 @@ def main(
...
         rpc_url=rpc_url,
-        initial_xdai_balance_per_agent=xdai_type(xdai_balance_per_agent),
+        initial_xdai_balance_per_agent=STARTING_AGENT_BALANCE,
     )
     reset_balances(
         rpc_url=rpc_url,
-        new_balance_agents_xdai=xdai_balance_per_agent,
-        new_balance_treasury_xdai=new_balance_treasury_xdai,
+        new_balance_agents_xdai=STARTING_AGENT_BALANCE,
+        new_balance_treasury_xdai=STARTING_TREASURY_BALANCE,
```
gabrielfior, line 44: IMO these constants should be configurable, and they aren't if defined inside PMAT.
I'm fine with any configurable option, e.g.
-> Define as ENV in the container (via Terraform)
-> Define as parameters in the script (like previously)

gabrielfior (approved):

prediction_market_agent/run_reset_game.py
kongzii: Can you expand on why? What advantage do you see in having this (and generally stuff like this) configurable through env?

In both cases mentioned, we still need to open up PR, approve, and merge.

I wouldn't care, but the motivation for me for the change was [because of this](https://github.com/gnosis/prediction-market-agent/blob/c09cfad5dac22d150f4f07afd1fcde4dcf3f3a15/prediction_market_agent/agents/microchain_agent/nft_treasury_game/tools_nft_treasury_game.py#L20). Some way to detect that something from the treasury was withdrawn.

If you have some way how to write it in another way, let's do that in follow up PR.

PR #758, opened 2025-03-18
Added hash checker for new safe tx

gabrielfior:


prediction_market_agent/agents/safe_guard_agent/guards/hash_checker.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def validate_safe_transaction_hash(
+    new_transaction: DetailedTransactionResponse,
+    new_transaction_safetx: SafeTx,
+    history: list[DetailedTransactionResponse],
+) -> ValidationResult:
+    """Function that checks if the safe_tx_hashes match."""
+
+    if not new_transaction.detailedExecutionInfo:
```
kongzii, line 18: Here, you can use `check_not_none` to get rid of mypy problems, because without detailedExecutionInfo, it should error out much sooner anyway: https://github.com/gnosis/prediction-market-agent/blob/e77d463430d9142b28b18883e5a685fab31a677b/prediction_market_agent/agents/safe_guard_agent/safe_api_utils.py#L104

prediction_market_agent/agents/safe_guard_agent/guards/hash_checker.py
```diff
@@ -0,0 +1,30 @@
...
+        return ValidationResult(ok=False, reason="Detailed execution info is missing.")
+
+    # Here we compare the safe_tx_hash delivered by the Safe client (see get_safe_detailed_transaction_info) and the
+    # safe constructed by safe-eth-py library (see the usage of safe.build_multisig_tx inside safe_tx_from_detailed_transaction).
+    # Note that the Safe Transaction Service offers a function (https://github.com/safe-global/safe-eth-py/blob/main/safe_eth/safe/api/transaction_service_api/transaction_service_api.py#L129)
+    # that also does a safe_tx_hash assertion. However, we choose to do it like this for more granular control.
+    safe_tx_hash_from_new_transaction = new_transaction.detailedExecutionInfo.safeTxHash
+    safe_tx_hash_from_built_tx = new_transaction_safetx.safe_tx_hash
+    if safe_tx_hash_from_new_transaction != safe_tx_hash_from_built_tx:
```
kongzii, line 27: I liked the idea from TG Group to verify hashes retrieved from multiple sources, but I'm thinking if we are getting any security check here?

SafeTX is build from the `DetailedTransactionResponse` [itself in here](https://github.com/gnosis/prediction-market-agent/blob/e77d463430d9142b28b18883e5a685fab31a677b/prediction_market_agent/agents/safe_guard_agent/safe_api_utils.py#L99), I guess if the hacker spoofed `DetailedTransactionResponse`, it would be in a such a way that SafeTX created out of it will be with the same hash?

However I'd still keep this check, why not! But at least one 3rd party hash source would be nice, is that possible somehow?

kongzii (approved):

PR #768, opened 2025-03-26
Safe Guard without signing + fix models

kongzii:
A few issues after this PR:

- https://github.com/gnosis/prediction-market-agent/issues/769
- https://github.com/gnosis/prediction-market-agent/issues/770

### DIFF
--- a/prediction_market_agent/agents/safe_guard_agent/app.py
+++ b/prediction_market_agent/agents/safe_guard_agent/app.py
@@ -1,17 +1,22 @@
 import nest_asyncio
 import streamlit as st
 from prediction_market_agent_tooling.config import APIKeys
+from prediction_market_agent_tooling.gtypes import ChecksumAddress
+from prediction_market_agent_tooling.tools.utils import check_not_none
 from web3 import Web3
 
 from prediction_market_agent.agents.safe_guard_agent import safe_api_utils
 from prediction_market_agent.agents.safe_guard_agent.safe_guard import (
     validate_safe_transaction,
 )
+from prediction_market_agent.agents.safe_guard_agent.safe_utils import check_if_owner
 from prediction_market_agent.tools.streamlit_utils import add_sink_to_logger
 
 nest_asyncio.apply()  # Required for streamlit to work with asyncio.
 st.set_page_config(layout="wide")
 add_sink_to_logger()
+
+keys = APIKeys()
 
 st.markdown(
     f"""# Safe Guard -- Fraud Detection Agent for Safe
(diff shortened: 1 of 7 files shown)


prediction_market_agent/agents/safe_guard_agent/safe_api_models/detailed_transaction_info.py
```diff
@@ -27,81 +29,88 @@ class TransferInfo(BaseModel):
     value: str
 
 
+class AddressInfo(BaseModel):
```
kongzii, line 32: For Pydantic models, they are mostly auto-generated based on transactions. After the option to allow validation of non-owner Safes, I used Martin's safe transactions to auto-fix these models. I think we don't have to worry about them too much atm as we have other issues for that.

prediction_market_agent/agents/safe_guard_agent/app.py
```diff
@@ -47,31 +53,89 @@
...
+do_execute = st.checkbox(
+    "Execute transaction if validated (possible only for Safe's owners)",
+    value=False,
+    disabled=not is_owner,
+)
+do_reject = st.checkbox(
+    "Reject transaction if not validated (possible only for Safe's owners)",
+    value=False,
+    disabled=not is_owner,
+)
```
gabrielfior, line 74: Do I understand correctly this is only displayed if agent is a signer? 
If e.g. we have the use case where a given proposed tx should be analyzed, then no reason to display this.

prediction_market_agent/agents/safe_guard_agent/app.py
```diff
@@ -47,31 +53,89 @@
...
+
+transaction = check_not_none(
+    next(
+        (
+            tx
+            for tx in queued_transactions + historical_transactions
+            if tx.id == transaction_id
+        ),
+        None,
+    )
```
gabrielfior, line 126: Sorry, why was the previous implementation not good enough? I didn't follow this change.

prediction_market_agent/agents/safe_guard_agent/app.py
kongzii: If he isn't the owner, the checkbox is disabled but visible. A way to let people know that if they do make him a signer, they can try this out too.

prediction_market_agent/agents/safe_guard_agent/app.py
kongzii: Previously, `transaction_id` was passed to the function, but because `queued` and `historical` transactions are two separate endpoints, it's not enough to pass only the id.

Also if we have the whole object here, why not just pass it instead of pass the ID + re-query the endpoint.

prediction_market_agent/agents/safe_guard_agent/safe_api_models/transactions.py
```diff
@@ -7,58 +7,61 @@
...
+    imitation: bool | None = None
 
 
-class Recipient(BaseModel):
+class Address(BaseModel):
     value: str
-    name: str | None
-    logoUri: str | None
+    name: str | None = None
+    logoUri: str | None = None
```
gabrielfior, line 20: Why do we need the logoUri for the processing logic?

kongzii: It just was in the responses, so auto-generation included it here.

PR #773, opened 2025-03-26
Added Coinflip seer agent

gabrielfior:


prediction_market_agent/agents/coinflip_agent/deploy_seer.py
```diff
@@ -0,0 +1,16 @@
...
+from prediction_market_agent_tooling.markets.agent_market import SortBy
+
+from prediction_market_agent.agents.coinflip_agent.deploy import DeployableCoinFlipAgent
+
+
+class DeployableSeerCoinFlipAgent(DeployableCoinFlipAgent):
+    n_markets_to_fetch = 5
+    get_markets_sort_by = SortBy.HIGHEST_LIQUIDITY
+    bet_on_n_markets_per_run = 2
+    same_market_trade_interval: TradeInterval = FixedInterval(timedelta(days=14))
```
gabrielfior, line 16: For betting on Seer in various markets and test out the integration, not focus on performance now.

kongzii: Why is a specific agent required?  It shouldn't be 🤔 

All agents should work on Seer by running them as `run_agent.py agent-name seer`.

But if the goal are these specific parameters, then maybe just change name to something like `DeployableCoinFlipAgentByHighestLiquidity`? (or anything else that isn't market specific if it isn't required.)

gabrielfior:
```
n_markets_to_fetch = 5
    get_markets_sort_by = SortBy.HIGHEST_LIQUIDITY
    bet_on_n_markets_per_run = 2
    same_market_trade_interval: TradeInterval = FixedInterval(timedelta(days=14))
```
these things are hardcoded, hence a new agent is Required.
Agree with name `highest_liquidity`, implementing.

kongzii (approved):

PR #775, opened 2025-03-28
Safe Guard: Type narrow and skip already processed transactions in th…

kongzii:
…e job

### DIFF
--- a/prediction_market_agent/agents/safe_guard_agent/safe_api_models/transactions.py
+++ b/prediction_market_agent/agents/safe_guard_agent/safe_api_models/transactions.py
@@ -136,6 +136,14 @@ class Transaction(BaseModel):
     txHash: str | None = None
 
 
+class TransactionWithMultiSig(Transaction):
+    executionInfo: MultisigExecutionInfo
+
+    @staticmethod
+    def from_tx(tx: Transaction) -> "TransactionWithMultiSig":
+        return TransactionWithMultiSig.model_validate(tx.model_dump())
+
+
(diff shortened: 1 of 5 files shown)


prediction_market_agent/agents/safe_guard_agent/safe_api_models/transactions.py
```diff
@@ -136,6 +136,14 @@ class Transaction(BaseModel):
     txHash: str | None = None
 
 
+class TransactionWithMultiSig(Transaction):
+    executionInfo: MultisigExecutionInfo
```
kongzii, line 140: Type narrowing the general Transaction into transaction with multisig, so we don't have to if-else it everywhere.

prediction_market_agent/agents/safe_guard_agent/safe_api_utils.py
```diff
@@ -74,12 +82,19 @@ def get_safe_queued_transactions(
...
     ).json()
     response_parsed = TransactionResponse.model_validate(response)
-    transactions = [
-        check_not_none(item.transaction)
-        for item in response_parsed.results
-        if is_valued_transaction_result(item, response_parsed.results)
+    return [r.transaction for r in response_parsed.results if r.transaction is not None]
+
+
+def get_safe_queue_multisig(
```
kongzii, line 88: Comparet do the standard `get_safe_queue`, this will return only multi sig transactions (including the narrowed model as seen above). Because only these (I hope 😄 ) makes sense for the agent to check.

PR #780, opened 2025-04-01
Formatting of Safe data for LLM

kongzii:
This is how LLM prompt looks like now; much better!

image

### DIFF
--- a/prediction_market_agent/agents/safe_guard_agent/guards/llm.py
+++ b/prediction_market_agent/agents/safe_guard_agent/guards/llm.py
@@ -96,7 +96,6 @@ def format_transaction(tx: DetailedTransactionResponse) -> str:
             if tx.executedAt
             else ""
         )
-        + f"Status: {tx.txStatus} | "
         + format_tx_data(tx.txData)
         if tx.txData
         else ""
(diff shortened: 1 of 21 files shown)


prediction_market_agent/agents/safe_guard_agent/safe_api_models/detailed_transaction_info.py
```diff
@@ -12,69 +12,28 @@
 )
 
 
-class Sender(BaseModel):
```
kongzii, line 15: Just noticed that these types were generated as duplicities, so removing them and using just one of them

prediction_market_agent/agents/safe_guard_agent/guards/llm.py
```diff
@@ -82,48 +89,179 @@ def validate_safe_transaction_llm(
...
+        formatted += format_settings_change_info(tx_info)
+
+    elif isinstance(tx_info, TransferTxInfo):
+        formatted += format_transfer_info(tx_info)
+
+    elif isinstance(tx_info, SwapOrderTxInfo):
+        formatted += format_swap_order_info(tx_info)
+
+    elif isinstance(tx_info, CustomTxInfo):
+        formatted += format_custom_info(tx_info)
```
gabrielfior, line 121: Would it be possible to add formatting to the classes themselves (e.g. `CreationTxInfo.format()` or similar) so that we can have just one call here and no type assertions?

kongzii: Sure looks nice. Refactored!

gabrielfior (approved):

PR #781, opened 2025-04-01
API for Safe Guard

kongzii:


prediction_market_agent/agents/safe_guard_agent/api.py
```diff
@@ -0,0 +1,66 @@
+import fastapi
```
kongzii, line 1: Not great to have this API in PMA, but I think the whole Safe Guard agent should have its own repository as it isn't for for PMA. So just keeping everything together so we can cut it out later at one piece.

prediction_market_agent/agents/safe_guard_agent/app.py
```diff
@@ -115,27 +115,15 @@ def get_safe_history_cached(
...
-        None,
-    )
-)
-
 
 def run_validation() -> None:
     validate_safe_transaction(
-        safe_address_checksum,
-        transaction,
+        check_not_none(transaction_id),
```
kongzii, line 121: Refactored validate_safe_transaction back to accept transaction id, because that is required in API endpoint.

prediction_market_agent/agents/safe_guard_agent/validation_result.py
```diff
@@ -4,3 +4,20 @@
 class ValidationResult(BaseModel):
     reason: str
     ok: bool
+
+
+class ValidationResultWithName(ValidationResult):
+    name: str
```
kongzii, line 10: Added the name field, because sometimes it isn't very informative without it, see:

image

PR #800, opened 2025-04-11
Validate all addresses from Safe transaction

kongzii:


prediction_market_agent/agents/safe_guard_agent/guards/goplus_.py
```diff
@@ -26,68 +27,72 @@ class GoPlusError(Exception):
...
     new_transaction: DetailedTransactionResponse,
     new_transaction_safetx: SafeTx,
+    all_addresses_from_tx: list[ChecksumAddress],
     history: list[DetailedTransactionResponse],
 ) -> ValidationResult | None:
-    addr = new_transaction_safetx.to
     data = _goplus_call(
         lambda: Token().token_security(
-            chain_id=f"{RPCConfig().chain_id}", addresses=[addr]
+            chain_id=f"{RPCConfig().chain_id}", addresses=all_addresses_from_tx
```
gabrielfior, line 35: Do they also use our convention for `chain_id`? I thought they used `xdai` somewhere as chain name, so just checking

gabrielfior (approved):

prediction_market_agent/agents/safe_guard_agent/guards/goplus_.py
kongzii: Yep they use the general convenction, 1 for ETH, 100 for Gnosis, etc. https://docs.gopluslabs.io/reference/response-details-9

PR #807, opened 2025-04-17
Test around signing and executing in Safe Guard

kongzii:


prediction_market_agent/agents/safe_guard_agent/safe_utils.py
```diff
@@ -103,27 +103,28 @@ def sign_or_execute(safe: Safe, tx: SafeTx, api_keys: APIKeys) -> None:
...
+        return None
     else:
-        logger.info("Threshold met, executing.")
+        logger.info(
+            f"Threshold {threshold} met with {len(tx.signers)} signs, executing."
+        )
         tx.call()
-        tx.execute(api_keys.bet_from_private_key.get_secret_value())
+        _, tx_params = tx.execute(api_keys.bet_from_private_key.get_secret_value())
+        return tx_params
```
gabrielfior, line 120: Why is the agent executing the transaction here? I would have expected the proposer to execute the transaction, the agent should be more like an "observer", no?

tests_integration_with_local_chain/conftest.py
```diff
@@ -41,12 +46,14 @@ def local_web3(
         fund_account_on_tenderly(
             tenderly_fork_rpc, [a.address for a in accounts], xDai(1000)
         )
+        set_env_rpc_url(tenderly_fork_rpc)
```
gabrielfior, line 49: Why is this needed now? Can't we continue passing `web3` instances around?

prediction_market_agent/agents/safe_guard_agent/safe_utils.py
kongzii: It was executing it before as well. Imagine you submit a transaction in Safe App, in the case the threshold is met, as a user, would you prefer to go back and execute it manually? So every few minutes, go back and check if you can execute it already. Or for the agent to execute it for you, if the threshold is met? I like the second option more.

tests_integration_with_local_chain/conftest.py
kongzii: If `RPCConfig` is initialised somewhere in the code, it didn't pick up the localchain rpc URL, but the production one from ENV.

image

prediction_market_agent/agents/safe_guard_agent/safe_utils.py
gabrielfior: I would not 100% trust the agent, as is the case for some slightly more paranoid users - but anyways, if this is the flow you prefer, works for me.

gabrielfior (approved):

PR #809, opened 2025-04-22
Retry for generate_and_set_image_for_market

kongzii:


prediction_market_agent/agents/replicate_to_omen_agent/image_gen.py
```diff
@@ -17,6 +18,10 @@
 
 
 @observe()
+@tenacity.retry(
```
kongzii, line 21: This started to error out, but all previously failed markets that I tried locally went OK. So I think OpenAI has problems with its image generation endpoint.

PR #810, opened 2025-04-22
Fix OFV Factuality validator

kongzii:


prediction_market_agent/agents/ofvchallenger_agent/ofv_models.py
```diff
@@ -7,7 +7,9 @@
     BeforeValidator(
         lambda v: (
             None
-            if str(v).lower().strip() in ("nothing to check.", "non-factual", "none")
+            if str(v).lower().strip() in ["none"]
+            or "non-factual" in str(v).lower()
```
kongzii, line 11: Last time LLM returned `"non-factual due to insufficient evidence"`, so trying to do it more universally now.

PR #813, opened 2025-04-23
Just warn in OFV on skipped markets

kongzii:


prediction_market_agent/agents/ofvchallenger_agent/deploy.py
```diff
@@ -161,8 +161,8 @@ def challenge_market(
             )
 
         if answer is None:
-            logger.error(
-                f"Failed to get factuality for market {market.url=}, question {market.question_title=}. Skipping."
+            logger.warning(
+                f"OFV didn't factcheck {market.url=}, question {market.question_title=}. Skipping."
```
kongzii, line 165: Only case where it returns None is this, that doesn't need to be logged as error:

image

PR #816, opened 2025-04-29
Go plus cache

kongzii:


prediction_market_agent/agents/safe_guard_agent/guards/goplus_.py
```diff
@@ -235,6 +241,46 @@ def _build_validation_result(malicious_reasons: list[str]) -> ValidationResult:
...
+def goplus_token_security(
+    address: ChecksumAddress, chain_id: int = RPCConfig().chain_id
+) -> dict[str, Any] | None:
+    # Used per-address instead of batching the call, so we can granualy cache the results and re-use across users.
+    data = _goplus_call(
+        lambda: Token().token_security(chain_id=f"{chain_id}", addresses=[address])
+    )
+    if data is None or not data.result:
+        return None
+    result: dict[str, Any] = data.result[address.lower()].to_dict()
```
kongzii, line 254: They return some objects (standard class, not pydantic, not typeddict) that aren't de/serializable. So I refactored it to work with dictionaries that can be cached using db_cache

gabrielfior: This works, but can't you pass a key arg like so?

```
@cachedmethod(lambda self: self.cache, key=partial(methodkey, method='pep'))
```

https://cachetools.readthedocs.io/en/latest/#cachetools.cachedmethod

PR #832, opened 2025-05-06
Safe Guard won't approve removal of itself from owners

kongzii:


prediction_market_agent/agents/safe_guard_agent/guards/agent.py
```diff
@@ -0,0 +1,43 @@
...
+@observe()
+def validate_do_not_remove_agent(
+    new_transaction: DetailedTransactionResponse,
+    new_transaction_safetx: SafeTx,
+    all_addresses_from_tx: list[ChecksumAddress],
+    history: list[DetailedTransactionResponse],
+) -> ValidationResult:
+    ok_result = ValidationResult(
+        ok=True, reason="Doesn't remove agent from the owner list."
+    )
```
gabrielfior, line 23: Let's assume there's a Safe with 2/3 threshold (alice, bob and safe guard).
Let's say Alice doesn't like the Safe Guard, and convinces Bob to remove it too.
Do I understand correctly that the Agent will NOT validate the transaction, however will not reject it as well, correct? I.e. it should be possible for 2 signers to remove the agent.

kongzii: Agent will create rejection tx but that's not end of the world, If you can meet threshold using other signers, you can always ignore the rejection tx and sign/execute the original one

prediction_market_agent/agents/safe_guard_agent/guards/agent.py
```diff
@@ -0,0 +1,43 @@
...
+
+    if not new_transaction.txData.dataDecoded:
+        return ok_result
+
+    if new_transaction.txData.dataDecoded.get("method") != "removeOwner":
+        return ok_result
+
+    if (
+        # Based on https://github.com/safe-global/safe-smart-account/blob/main/contracts/base/OwnerManager.sol#L73.
+        new_transaction.txData.dataDecoded["parameters"][1]["value"].lower()
```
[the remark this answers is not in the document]
kongzii: In the decoded parameters, there is a list of arguments passed to the function. Can't do anything about that. But as long as contract doesn't change, this will work. And there is a test for it.

gabrielfior (approved):

Issue #712, opened 2025-02-25
Store logsprobs in langfuse


kongzii:
Whenever possible, store logprobs for LLMs in Langufse

TS9001:
Currently it is impossible to return logprobs from PydanticAI, fortunately there is [PR](https://github.com/pydantic/pydantic-ai/issues/1228 ) already for this 
I propose to follow PR proposed functionality by overriding current OpenAIModel and storing logprobs to [ProphetAgent](https://github.com/agentcoinorg/predictionprophet/blob/main/prediction_prophet/autonolas/research.py#L1167)  response and consecutively to the API. This will mean that we will get the functionality with minimal ammount of changes and logprobs can be enabled by providing  parameter `extra_body = {"logprobs": True, "top_logprobs": 3}` to[ Model settings](https://github.com/pydantic/pydantic-ai/blob/dd22595464ea430d9fcea388e506bf0a1697a9a4/pydantic_ai_slim/pydantic_ai/settings.py#L7) during model creation.


PR for [PAM](https://github.com/gnosis/prediction-market-agent/pull/833)
PR for [ProphetAgent](https://github.com/agentcoinorg/predictionprophet/pull/130)

PR #833, opened 2025-05-06
Modify OAI agent so it can parse logprobs

TS9001:
Linked to https://github.com/gnosis/prediction-market-agent/issues/712


kongzii (approved):

prediction_market_agent/agents/logprobs_oai_model.py
```diff
@@ -0,0 +1,74 @@
...
+    TextPart,
+    ToolCallPart,
+)
+from pydantic_ai.models.openai import OpenAIModel
+from pydantic_ai.providers.openai import OpenAIProvider
+
+
+@dataclass
+class LogProbsModelResponse(ModelResponse):
+    vendor_details: dict[str, Any] | None = field(default=None, repr=False)
```
gabrielfior, line 20: Why dataclass and not pydantic.BaseModel? Asking because we use BaseModel by default, so just curious =)

TS9001: Unfortunatelly here I am following what is inside of [PydanticAI](https://github.com/pydantic/pydantic-ai/blob/f18e65dc09a965d429a355ccd63ed43561f2d676/pydantic_ai_slim/pydantic_ai/messages.py#L535)

PR #837, opened 2025-05-09
Get rid of _state error logs

kongzii:


TS9001 (approved):
:heavy_check_mark:  LGTM!

PR #842, opened 2025-05-12
Fix base url for openai provider

kongzii:


prediction_market_agent/tools/openai_utils.py
```diff
@@ -22,6 +22,7 @@ def get_openai_provider(
     return OpenAIProvider(
         openai_client=AsyncOpenAI(
             api_key=api_key.get_secret_value(),
+            base_url=base_url,
```
kongzii, line 25: For some reason, `base_url` needs to be provided here and also 3 lines down.. Because of this, OpenRouter-based agents doesn't work. My bad I didn't try them out after these changes (OpenAI-based agents work fine, as for them,`base_url` defaults to the correct one)

TS9001 (approved):

PR #845, opened 2025-05-13
Script for resolution of unresolved markets

TS9001:


scripts/resolve_markets.py
```diff
@@ -0,0 +1,51 @@
...
+            logger.error(f"Failed to retrieve market {market_id}: {e}")
+
+    for market in markets:
+        logger.info(
+            f"Challenging market {market.question.id=} - {market.question.title=} - {market.question.outcomes=}"
+        )
+        logger.info(f"Market creation timestamp: {market.creation_datetime}")
+        logger.info(f"Market close time: {market.close_time}")
+        if market.close_time and not market.is_resolved_with_valid_answer:
+            current_time = datetime.now(timezone.utc)
```
kongzii, line 39: Use

```
from prediction_market_agent_tooling.tools.utils import utcnow
```

here please, we use it everywhere to be consistent with UTC timezone and it returns a specialized type-safe DatetimeUTC type.

scripts/resolve_markets.py
```diff
@@ -0,0 +1,51 @@
...
+
+from prediction_market_agent.agents.ofvchallenger_agent.deploy import OFVChallengerAgent
+from prediction_market_agent.utils import APIKeys
+
+MARKET_IDS: list[HexAddress] = [
+    HexAddress(HexStr("000000000000000000000000000000000000000000")),
+]
+
+
+def main() -> None:
```
kongzii, line 19: You used `typer.run(main)` which is cool, that allows you to do:

```
def main(market_id: list[str]) -> None:
```

and then script can be used as

```
python scripts/resolve_markets.py id-1 id-2 ...
```

so no hard-coding of market ids is necessary. But you will need to convert them to proper type using

```
from web3 import Web3
Web3.to_checksum_address(addr)
```

kongzii (commented):
Small codebase-related nits, otherwise LGTM!

scripts/resolve_markets.py
TS9001: Fixed, thank you!

scripts/resolve_markets.py
```diff
@@ -0,0 +1,53 @@
...
+from prediction_market_agent_tooling.markets.omen.omen_subgraph_handler import (
+    OmenSubgraphHandler,
+)
+from prediction_market_agent_tooling.tools.utils import utcnow
+from web3 import Web3
+
+from prediction_market_agent.agents.ofvchallenger_agent.deploy import OFVChallengerAgent
+from prediction_market_agent.utils import APIKeys
+
+MARKET_IDS: list[str] = ["000000000000000000000000000000000000000000"]
```
kongzii, line 16: Why this as default argument? What happens if it's executed with this? 🤔

kongzii (approved):

PR #847, opened 2025-05-14
Nicer demo frontend for safe guard

kongzii:
Summary is now generated and only that is shown by default:

image

But you can click on Show details to see them:

image

But it will show only relevant stuff; no unnecessary logs are shown now.

If you have something in mind that you'd like to see there, lmk.


prediction_market_agent/agents/safe_guard_agent/app.py
```diff
@@ -10,8 +10,7 @@
 
 nest_asyncio.apply()  # Required for streamlit to work with asyncio.
 st.set_page_config(layout="wide")
-add_sink_to_logger()
-
+add_sink_to_logger(explicit=True)
```
kongzii, line 13: If explicit=True is used, then only logs with the extra attribute `streamlit=True` will be printed in Streamlit app. Eg:

```
logger.info("This is not visible on stremalit")
logger.info("But this is", streamlit=True)
```

It's handy if you want fine-grained control over the streamlit app, but don't want to get rid of handy logs for production.

prediction_market_agent/agents/safe_guard_agent/guards/agent.py
```diff
@@ -11,33 +14,44 @@
...
-        # Based on https://github.com/safe-global/safe-smart-account/blob/main/contracts/base/OwnerManager.sol#L73.
-        new_transaction.txData.dataDecoded["parameters"][1]["value"].lower()
-        != APIKeys().bet_from_address.lower()
-    ):
-        return ok_result
-
-    return ValidationResult(
-        ok=False, reason="Agent can not accept removal of itself from the owners."
-    )
+class DoNotRemoveAgent(AbstractGuard):
```
kongzii, line 17: Github shows these as big chunks of modifications -- but in reality, it's just indenting and adding the name and description properties.

pyproject.toml
```diff
@@ -46,7 +46,7 @@ sqlmodel = "^0.0.22"
 farcaster = { git = "https://github.com/gabrielfior/farcaster-py.git", branch = "update-eth-account" }
 streamlit-extras = "^0.4.2"
 tweepy = "^4.14.0"
-streamlit = "^1.40.0"
+streamlit-nightly = "1.45.2.dev20250513" # TODO: Switch to `streamlit = "^1.45.2"` once they release it.
```
kongzii, line 49: The version that supports nested expanders isn't released yet, but this works fine.

prediction_market_agent/agents/safe_guard_agent/guards/agent.py
```diff
@@ -11,33 +14,44 @@
...
+            description=self.description,
+            ok=True,
+            reason="The transaction does not remove the agent from owners.",
+        )
+
+        if not new_transaction.txData:
+            return ok_result
+
+        if not new_transaction.txData.dataDecoded:
+            return ok_result
```
gabrielfior, line 40: Probably a silly question - any reason for combining `not a or not b or ...`?

prediction_market_agent/agents/safe_guard_agent/guards/llm.py
```diff
@@ -23,49 +27,61 @@
...
+        history: list[DetailedTransactionResponse],
+    ) -> ValidationResult:
+        # Get the latest ones.
+        history = sorted(
+            history, key=lambda x: x.executedAt or float("-inf"), reverse=True
+        )[:HISTORY_LIMIT]
+
+        agent = Agent(
+            OpenAIModel(
+                "gpt-4o",
```
gabrielfior, line 54: I think I asked before, but - did you try using a cheaper model? Maybe this can be investigated in the evals, if gpt-4o-mini is too bad, otherwise we could use that (or even make the model a dropdown choice on the app)

prediction_market_agent/agents/safe_guard_agent/safe_utils.py
```diff
@@ -108,13 +109,14 @@ def sign_or_execute(safe: Safe, tx: SafeTx, api_keys: APIKeys) -> TxParams | Non
     threshold = safe.retrieve_threshold()
 
     if threshold > len(tx.signers):
-        logger.info("Threshold not met yet, just adding a sign.")
+        logger.info("Threshold not met yet, posting a signature.", stremalit=True)
```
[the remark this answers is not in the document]
gabrielfior: Good catch @coderabbitai

prediction_market_agent/agents/safe_guard_agent/safe_utils.py
```diff
@@ -108,13 +109,14 @@ def sign_or_execute(safe: Safe, tx: SafeTx, api_keys: APIKeys) -> TxParams | Non
...
         api = TransactionServiceApi(EthereumNetwork(RPCConfig().chain_id))
```
gabrielfior, line 113: I think we should also analyze Ethereum transactions, since they probably are quite similar.

I see that in `get_safe_history` (https://github.com/gnosis/prediction-market-agent/blob/a8fe16669550a852ceff92f5a5e8a3903dc2d8e8/prediction_market_agent/agents/safe_guard_agent/safe_api_utils.py#L142) the `chain_id` parameter is being fetched from .env.
Would it be too much work to add a dropdown to Streamlit allowing Ethereum (chain_id 1) to also be fetched?

prediction_market_agent/tools/streamlit_utils.py
```diff
@@ -47,57 +49,78 @@ def check_required_api_keys(required_keys: list[str]) -> None:
...
-    message = streamlit_escape(record["message"])
+        # Ignore certain messages that aren't interesting for Streamlit user, but are in the production logs.
+        if any(x in message for x in [DB_CACHE_LOG_PREFIX]):
+            return
 
-    # Ignore certain messages that aren't interesting for Streamlit user, but are in the production logs.
-    if any(x in message for x in [DB_CACHE_LOG_PREFIX]):
-        return
+        if explicit and STREAMLIT_SINK_EXPLICIT_FLAG not in record["extra"]:
+            return
```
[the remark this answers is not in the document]
kongzii: That's fair, implemented

prediction_market_agent/agents/safe_guard_agent/guards/goplus_.py
```diff
@@ -13,229 +13,272 @@
...
-        if result is None:
-            continue
+    @observe(name="validate_safe_transaction_goplus_token_security")
+    def validate(
+        self,
+        new_transaction: DetailedTransactionResponse,
+        new_transaction_safetx: SafeTx,
+        all_addresses_from_tx: list[ChecksumAddress],
+        history: list[DetailedTransactionResponse],
+    ) -> ValidationResult | None:
```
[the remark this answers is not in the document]
kongzii: Also good catch! Fixed

prediction_market_agent/agents/safe_guard_agent/guards/goplus_.py
```diff
@@ -13,229 +13,272 @@
...
+class GoPlusTokenSecurity(AbstractGuard):
+    name = "Token security"
+    description = "This guard checks the security of tokens involved in the transaction using GoPlus API."
+
+    @observe(name="validate_safe_transaction_goplus_address_security")
+    def validate(
+        self,
+        new_transaction: DetailedTransactionResponse,
+        new_transaction_safetx: SafeTx,
+        all_addresses_from_tx: list[ChecksumAddress],
```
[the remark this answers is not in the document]
kongzii: Again cool! Fixed

prediction_market_agent/agents/safe_guard_agent/guards/agent.py
kongzii: Not really, I'm fine with that 😄

prediction_market_agent/agents/safe_guard_agent/guards/llm.py
kongzii: Previously it was `o1`, so this is the cheaper one 😄 I wouldn't go as cheap as possible here. Traffic is low, and it needs to be at least somewhat clever. This has ~25k input tokens and ~100 output tokens, that's ~$0.127 per run.

Agree with evals, but I'm afraid we will need more expensive models, not cheaper.

prediction_market_agent/agents/safe_guard_agent/safe_utils.py
kongzii: No idea what will happen if we change chain id in the environment, we never did that so far 😄 But that's a problem for another day: https://github.com/gnosis/prediction-market-agent/issues/848

gabrielfior (approved):

PR #852, opened 2025-05-19
SG for ETH

kongzii:


prediction_market_agent/agents/safe_guard_agent/app.py
```diff
@@ -22,15 +23,24 @@
...
-    st.Page(demo_page.demo_page, title="Demo", url_path="demo"),
-    st.Page(api_page.api_page, title="API", url_path="api"),
-    st.Page(agents_page.agents_page, title="Agents", url_path="agents"),
+    st.Page(demo_page.get_demo_page(chain_id=chain_id), title="Demo", url_path="demo"),
+    st.Page(api_page.get_api_page(chain_id=chain_id), title="API", url_path="api"),
+    st.Page(
+        agents_page.get_agents_page(chain_id=chain_id),
+        title="Agents",
+        url_path="agents",
+    ),
```
gabrielfior, line 43: I don't like this hard separation too much - in theory, we could easily support all 19 networks where the Transaction Service operates (https://docs.safe.global/advanced/smart-account-supported-networks?service=Transaction+Service&version=v1.4.1).
In that case, we would have 19 pages in the app, which wouldn't be very user-friendly.
I would expect instead to have 1 page (with the chain dropdown) that simply is passed as a parameter to the underlying function.

prediction_market_agent/agents/safe_guard_agent/streamlit_pages/agents_page.py
```diff
@@ -22,7 +26,11 @@ def agents_page() -> None:
 
 Agent will also send you a message to your Safe with the result of the validation.
 """
-    )
+        )
+
+        agent_address = Config().SAFE_GUARD_AGENT_ADDRESS
+        st.markdown(
+            f"- `{agent_address}` - https://gnosisscan.io/address/{agent_address}"
+        )
```
gabrielfior, line 34: Should it be hardcoded to `gnosis scan` or is it chain-dependent?

prediction_market_agent/agents/safe_guard_agent/app.py
kongzii: > I would expect instead to have 1 page (with the chain dropdown) that simply is passed as a parameter to the underlying function.

But that's exactly how it is right now..

prediction_market_agent/agents/safe_guard_agent/streamlit_pages/agents_page.py
kongzii: True! Fixed

prediction_market_agent/agents/safe_guard_agent/app.py
gabrielfior:
```
pages = [
    st.Page(demo_page.get_demo_page(chain_id=chain_id), title="Demo", url_path="demo"),
    st.Page(api_page.get_api_page(chain_id=chain_id), title="API", url_path="api"),
    st.Page(
```
Ow I thought this meant having 1 page per chain. My bad for not following everything Streamlit offers. Closing this.

gabrielfior (approved):

Issue #811, opened 2025-04-22
Deploy agent from Tomas homework into production


kongzii:
The agent is using PMAT so it should be more or less straightforward.

1. Add an enum into RunnableAgent in https://github.com/gnosis/prediction-market-agent/blob/main/prediction_market_agent/run_agent.py
2. Create mapping in RUNNABLE_AGENTS
3. Try that `python prediction_market_agent/run_agent.py the-name-you-chosen omen` works
4. Create a new private key for him, save it in 1password and also GCP secrets
5. Create new OpenAI key or whatever it's using, also add to his GCP secret (where possible, we have key per agent so we can identify if they go rouge)
6. Copy pasta existing agent in infrastructure repo and edit accordingly

TS9001:
Implemented PerplexityClient in https://github.com/gnosis/prediction-market-agent-tooling/pull/700

TS9001:
Agent code and deployment in https://github.com/gnosis/prediction-market-agent/pull/857

TS9001:
Infrastructure PR https://github.com/gnosis/gnosisai-infrastructure/pull/155

TS9001:
Agent was deployed as logprobs agent.

PR #857, opened 2025-05-22
Slightly improved homework agent

TS9001:
https://github.com/gnosis/prediction-market-agent/issues/811

### DIFF
--- a/prediction_market_agent/agents/logprobs_agent/deploy.py
+++ b/prediction_market_agent/agents/logprobs_agent/deploy.py
@@ -1,0 +1,261 @@
+import asyncio
+import os
+import json
+from typing import Any, Literal
+from gpt_researcher import GPTResearcher
+from prediction_market_agent_tooling.deploy.agent import DeployableTraderAgent
+from prediction_market_agent_tooling.markets.agent_market import AgentMarket
+from prediction_market_agent_tooling.markets.data_models import ProbabilisticAnswer
+from prediction_market_agent_tooling.tools.langfuse_ import observe
+from pydantic_ai import Agent
+from pydantic_ai.agent import AgentRunResult
+from pydantic_ai.models.openai import OpenAIModel
+from pydantic_ai.settings import ModelSettings
+from prediction_market_agent_tooling.loggers import logger
+from pydantic import BaseModel
+from prediction_market_agent.tools.openai_utils import get_openai_provider
+from prediction_market_agent.utils import APIKeys
+from prediction_market_agent.agents.logprobs_oai_model import LogProbsOpenAIModel
+from prediction_market_agent_tooling.tools.perplexity.perplexity_search import perplexity_search
+from prediction_market_agent_tooling.tools.perplexity.perplexity_models import PerplexityResponse
+from prediction_market_agent.tools.web_scrape.structured_summary import web_scrape_structured_and_summarized
+from prediction_market_agent_tooling.logprobs_parser import LogprobsParser, LogprobDetail, FieldLogprobs
+from prediction_market_agent_tooling.markets.markets import MarketType
+from prediction_market_agent_tooling.markets.markets import get_binary_markets
+from concurrent.futures import ThreadPoolExecutor
+from prediction_market_agent_tooling.gtypes import Probability
+
+class DecidabilityResponse(BaseModel):
+    rationale: str
+    answer: Literal["YES", "NO"]
+
+class PredictionResponse(BaseModel):
+    rationale: str
+    p_yes: float
+
+PERPLEXITY_QUERY = """
+Your goal is to provide assesment of sources for following question: {market_question}?
+
+* Find relevant information. Every time evaluate quality and relevance of your soruces.
+* Don't answer the question, just provide assesment of quality and relevance of ALL given sources.
+* Each assesment MUST start with LINK OF THE SOURCE so i can pair it with given source later.
+* In your assesment also provide critique of given source - website etc..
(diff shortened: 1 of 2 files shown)


prediction_market_agent/agents/logprobs_agent/deploy.py
```diff
@@ -0,0 +1,261 @@
...
+
+        summaries = self._process_links_parallel(market.question, research_results.citations, max_workers=10)
+        logger.info(f"Summaries of sources: {summaries}")
+
+        if self.use_solvability_score and self.processed_markets < self.max_processed_markets:
+            solvability_score = self._calculate_solvability_score(market.question, research_results.content, summaries)
+            logger.info(f"Solvability score: {solvability_score}")
+            if solvability_score is not None and solvability_score < self.min_solvability_score:
+                logger.info(f"Skipping market, min solvability {self.min_solvability_score} not reached")
+                self.bet_on_n_markets_per_run += 1
```
kongzii, line 133: `bet_on_n_markets_per_run` is part of PMAT framework, not something that should be modified in the agent implementations, see https://github.com/gnosis/prediction-market-agent-tooling/blob/2ff3e2d7081c257bacd67e6c936cf8ea9b323172/prediction_market_agent_tooling/deploy/agent.py#L470

It will break the loop if agent made enough of bets (and it doesn't count if `None` was returned, in that case, bet isn't made).

prediction_market_agent/agents/logprobs_agent/deploy.py
```diff
@@ -0,0 +1,261 @@
...
+                return None
+
+            self.processed_markets += 1
```
kongzii, line 136: This is useful variable to have! You could promote this to PMAT itself, made it `self.processed` and then you can just use it in your agent here: https://github.com/gnosis/prediction-market-agent-tooling/blob/2ff3e2d7081c257bacd67e6c936cf8ea9b323172/prediction_market_agent_tooling/deploy/agent.py#L457

prediction_market_agent/agents/logprobs_agent/deploy.py
TS9001: Yep, makes absolute sense - gone - actually it is not needed at all if None does not count (I probably misread the code :man_facepalming: )

prediction_market_agent/agents/logprobs_agent/deploy.py
TS9001: not needed, see above

prediction_market_agent/agents/logprobs_agent/deploy.py
```diff
@@ -0,0 +1,286 @@
...
+        prediction = self.research_agent.run_sync(
+            PREDICTION_QUERRY.format(
+                market_question=market_question,
+                critique=critique,
+                summary=crawl_summaries,
+            )
+        )
+        raw_logprobs = self.get_logprobs(prediction)
+        if raw_logprobs is None:
+            return None
```
kongzii, line 256: Shouldn't this be raised?

If logprobs aren't available at all, sounds as a bug on either our side or openai side.

prediction_market_agent/agents/logprobs_agent/deploy.py
```diff
@@ -0,0 +1,286 @@
...
+        # Get probabilities for "YES" and "NO" tokens in the answer field
+        probs = self._extract_token_probabilities(
+            raw_logprobs=raw_logprobs,
+            model_type=DecidabilityResponse,
+            field_key="answer",
+            expected_tokens=["YES", "NO"],
+        )
+
+        yes_prob = probs.get("yes", 0)
+        no_prob = probs.get("no", 0)
```
kongzii, line 197: Shouldn't this be raised if they are both 0?

Because if that happens, it means the agent will be `1`, even if in reality he had no clue about the predicted probability.

kongzii: And if `expected_tokens=["YES", "NO"],`, how come here you are getting lowercased values?

prediction_market_agent/agents/logprobs_agent/deploy.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+class DecidabilityResponse(BaseModel):
+    rationale: str
+    answer: Literal["YES", "NO"]
+
+
+class PredictionResponse(BaseModel):
+    rationale: str
+    p_yes: float = Field(ge=0, le=1)
```
kongzii, line 36: Is there any theory behind why logprobs are used for `DecidabilityResponse`, but not for `PredictionResponse`? 🤔

prediction_market_agent/agents/logprobs_agent/deploy.py
TS9001: Reworked so it works in more concise manner, removed wrong to_lower.

prediction_market_agent/agents/logprobs_agent/deploy.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+def clean_json_response(response_str: str) -> PredictionResponse:
+    response_str = response_str.strip().replace("\n", " ")
+    response_str = " ".join(response_str.split())
+    start_index = response_str.find("{")
+    end_index = response_str.rfind("}")
+    response_str = response_str[start_index : end_index + 1]
+
+    return PredictionResponse.model_validate_json(response_str)
```
TS9001, line 288: Changed this to have same approach as there is in prophet - it is safer

PR #873, opened 2025-05-30
Rename Safe Guard to Safe Watch

kongzii:


TS9001 (approved):
> _Probably OK (if you renamed everything :D ) One question - why did you decided not to rename the package??_

**Disregard this comment some mess-up on my side - everything is correct**

prediction_market_agent/agents/safe_watch_agent/old_app.py
```diff
@@ -0,0 +1,9 @@
+import streamlit as st
+
+st.markdown(
+    """
+# Safe Guard is now Safe Watch! 
+            
+Please head to https://safewatch.ai.gnosisdev.com.
+"""
+)
```
TS9001, line 9: Would it be possible to redirect user automatically let's say after 5 seconds? Probably more user-friendly (only if you think it makes sense [probably does not])

PR #874, opened 2025-05-30
Fix Prophet if prediction is None

kongzii:


prediction_market_agent/agents/prophet_agent/deploy.py
```diff
@@ -67,8 +67,12 @@ def answer_binary_market(self, market: AgentMarket) -> ProbabilisticAnswer | Non
             logger.info(
                 f"Answering '{market.question}' with '{prediction.outcome_prediction}'."
             )
-            outcome_prediction = check_not_none(prediction.outcome_prediction)
```
TS9001, line 70: Probably now redundant import

TS9001 (approved):

PR #888, opened 2025-06-05
Cyvers for SWA

kongzii:


prediction_market_agent/agents/safe_watch_agent/watchers/cyvers.py
```diff
@@ -0,0 +1,122 @@
...
+                continue
+
+            logger.info(
+                f"{self.name} response for address {addr}:\n\n{dict_to_point_list(result_for_addr)}",
+                streamlit=True,
+            )
+
+            if (risk_score := result_for_addr["data"]["risk_score"].lower()) != "low":
+                malicious_reasons.append(
+                    f"Address {addr} has high risk score {risk_score}."
```
TS9001, line 54: Nitpick: I would change this to something like  `f"Address {addr} has risk score {risk_score}."` Otherwise you can have messages like "Address 0x123456 has high risk score medium." - this might be confusing

prediction_market_agent/agents/safe_watch_agent/watchers/cyvers.py
```diff
@@ -0,0 +1,122 @@
...
+
+        risk_level = simulation_response["data"]["risk_level"]
+        risk_description = simulation_response["data"]["risk_description"]
+
+        if risk_level.lower() != "low":
+            malicious_reasons.append(
+                f"Safe TX has a high risk level {risk_level}, because of {risk_description}."
+            )
+
+        return _build_validation_result(self.name, self.description, malicious_reasons)
```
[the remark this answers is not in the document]
TS9001: Probably agree with rabbit

prediction_market_agent/agents/safe_watch_agent/watchers/cyvers.py
```diff
@@ -0,0 +1,122 @@
...
+def cyvers_address_reputation(address: str) -> dict[str, Any]:
+    url = f"https://vigilens-api.cyvers.ai/v1/risk-score/{address}"
+    headers = {
+        "accept": "application/json",
+        "authorization": f"Bearer {APIKeys().cyvers_api_key.get_secret_value()}",
+    }
+    response = requests.get(url, headers=headers)
+    response.raise_for_status()
+    response_json: dict[str, Any] = response.json()
+    return response_json
```
TS9001, line 100: So can it be None or not? Also see rabit comment...

prediction_market_agent/agents/safe_watch_agent/watchers/cyvers.py
```diff
@@ -0,0 +1,122 @@
...
+    }
+    headers = {
+        "accept": "application/json",
+        "content-type": "application/json",
+        "authorization": f"Bearer {APIKeys().cyvers_api_key.get_secret_value()}",
+    }
+    response = requests.post(url, json=payload, headers=headers)
+    response.raise_for_status()
+    response_json: dict[str, Any] = response.json()
+    return response_json
```
TS9001, line 122: Same as above.

TS9001 (commented):
Hi @kongzii, Great work! just some nitpicks

prediction_market_agent/agents/safe_watch_agent/watchers/cyvers.py
kongzii: This is interesting, I thought it can not but you made me double check 😄 In theory, if API returns just `"null"` then it will be `None`:

Screenshot by Dropbox Capture

I think in practice, this shouldn't happen, as I added `response.raise_for_status()` above. But agree that we shouldn't just blindly trust the API 😄 As they could return `"null"` with 200 response. 🤷  So I added the check on both places:

Screenshot by Dropbox Capture

TS9001 (approved):

PR #894, opened 2025-06-11
Moved ProphetAgent benchmarks from PMAT

TS9001:
* Agents for prophet need to be moved to PMA because of cyclic dependency

See also:
https://github.com/gnosis/prediction-market-agent-tooling/pull/730

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@ logs
 .mypy_cache
 .pytest_cache
 .venv
+.history/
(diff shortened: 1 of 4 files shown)


kongzii (approved):
This is all just copy-pasta of already reviewed code, right? 😄

TS9001:
> This is all just copy-pasta of already reviewed code, right? 😄

exactly

Issue #908, opened 2025-06-23
Add support for Scalar markets for Seer


TS9001:
We want to add support for [scalar](https://seer-2.gitbook.io/seer/prediction-market-basics/scalar-markets) markets to [seer](https://seer.pm/).

Consider addition of [Scalar](https://app.seer.pm/markets/100/the-price-of-bitcoin-at-the-start-of-2025-usd-3/?outcome=UP) and [MultiScalar](https://app.seer.pm/markets/100/how-many-seats-in-frances-national-assembly-will-party-bloc-win-in-the-next-fren?outcome=Left+%28UG%2C+LFI%2C+SOC%2C+COM%2C+VEC%2C+RDG%2C+DVG%2C+EXG%2C+%E2%80%A6%29) markerts.

PR #910, opened 2025-06-23
Added basis for prophet scalar agent

TS9001:
Solves https://github.com/gnosis/prediction-market-agent/issues/908

Depends on https://github.com/gnosis/prediction-market-agent-tooling/pull/738 and https://github.com/agentcoinorg/predictionprophet/pull/136


prediction_market_agent/agents/prophet_agent/deploy.py
```diff
@@ -214,6 +216,114 @@ def load(self) -> None:
...
+                    provider=get_openai_provider(api_key=api_keys.openai_api_key),
+                ),
+                model_settings=ModelSettings(temperature=0.0),
+            ),
+            include_reasoning=True,
+            logger=logger,
+        )
+
+    def answer_binary_market(self, market: AgentMarket) -> ProbabilisticAnswer | None:
+        logger.warning("Binary markets are not supported for scalar agents for now")
```
kongzii, line 249: Then it can be just left un-implemented right?

Agent shouldn't fetch and try to process markets that he can not.

See how [def fetch_categorical_markets(self) -> bool:](https://github.com/gnosis/prediction-market-agent-tooling/blob/2d50d65a630b1683f921b3b2654fac535d51791a/prediction_market_agent_tooling/deploy/agent.py#L315) is done 🙏

prediction_market_agent/agents/prophet_agent/deploy.py
```diff
@@ -214,6 +216,114 @@ def load(self) -> None:
...
+        return None
+
+    def answer_categorical_market(
+        self, market: AgentMarket
+    ) -> CategoricalProbabilisticAnswer | None:
+        logger.warning(
```
kongzii, line 255: Similarly here 🙏

prediction_market_agent/agents/prophet_agent/deploy.py
```diff
@@ -214,6 +216,114 @@ def load(self) -> None:
...
+    #     return KellyBettingStrategy(
+    #         max_bet_amount=get_maximum_possible_bet_amount(
+    #             min_=USD(1),
+    #             max_=USD(5),
+    #             trading_balance=market.get_trade_balance(APIKeys()),
+    #         ),
+    #         max_price_impact=0.7,
+    #     )
+
+    def load(self) -> None:
```
kongzii, line 301: `DeployableTraderAgentERScalar` already defines this load method, can this be removed?

prediction_market_agent/agents/top_n_oai_model.py
```diff
@@ -0,0 +1,107 @@
...
+
+        try:
+            return await self.client.chat.completions.create(
+                model=self._model_name,
+                messages=openai_messages,
+                n=self._n,
+                parallel_tool_calls=model_settings.get(
+                    "parallel_tool_calls", NOT_GIVEN
+                ),
+                tools=tools or NOT_GIVEN,
```
gabrielfior, line 62: Sorry, why do we need all this boilerplate code? It seems a lot of work to simply get the logprobs, or do I misunderstand this?

See langchain
```
from langchain_openai import ChatOpenAI
llm = ChatOpenAI(model="gpt-4o-mini").bind(logprobs=True)
msg = llm.invoke(("human", "how are you today"))
msg.response_metadata["logprobs"]["content"][:5]
```
PydanticAI also has something similar https://ai.pydantic.dev/api/models/openai/#pydantic_ai.models.openai.OpenAIModelSettings

prediction_market_agent/agents/prophet_agent/deploy.py
TS9001: Yes exactly, gone!

prediction_market_agent/agents/prophet_agent/deploy.py
TS9001: Gone as well

prediction_market_agent/agents/prophet_agent/deploy.py
TS9001: Technically it has different temperature but yes, gone from ERScalar

prediction_market_agent/agents/top_n_oai_model.py
TS9001: This is not for log-probs but top-n competitions returning. Unfortunately Lang fuse currently hard-codes 1 response in competitions. In this case we have to overload whole method and wait until they implement it in issue mentioned in comment.

Using top-n allows us to take value from prob distribution of n tries without having to sent input tokens (and pay for them) several times.

prediction_market_agent/agents/prophet_agent/deploy.py
kongzii: Ah I didn't notice that! Up to you if you want to use different temperature here of course.

PR #912, opened 2025-06-27
906 replicate polymarket markets on omen with crc as base token

gabrielfior:


pyproject.toml
```diff
@@ -33,7 +33,7 @@ poetry = "^1.7.1"
 poetry-plugin-export = "^1.6.0"
 functions-framework = "^3.5.0"
 cron-validator = "^1.0.8"
-prediction-market-agent-tooling = "^0.66.0"
+prediction-market-agent-tooling = {git = "https://github.com/gnosis/prediction-market-agent-tooling.git", rev = "gabriel/fix-polymarket-replicate"}
```
gabrielfior, line 36: Will be changed after PMAT PR is merged

prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
```diff
@@ -53,22 +55,21 @@ def omen_replicate_from_tx(
     api_keys: APIKeys,
     market_type: MarketType,
     n_to_replicate: int,
-    initial_funds: USD,
+    initial_funds: USD | CollateralToken,
```
gabrielfior, line 58: We also allow the `initial_funds` quantity to be specified on a per-settings way, which is given in collateral units.

prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
```diff
@@ -132,17 +133,28 @@ def omen_replicate_from_tx(
...
         if is_invalid(market.question):
             logger.info(
-                f"Skipping `{market.question}` because it seems to be an invalid question."
+                f"Skipping `{market.question}` was marked as invalid. Trying to rephrase and make it valid."
             )
-            continue
+            new_question = rephrase(market.question)
+            logger.info(f"Rephrased `{market.question}` to `{new_question}`.")
+
+            # Reinsert at the beginning of the list to reprocess with the new question
```
kongzii, line 141: What do you mean by reinserting to beginning of list?

prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
```diff
@@ -132,17 +133,28 @@ def omen_replicate_from_tx(
...
+                market.question = new_question
 
         if not is_predictable_binary(market.question):
             logger.info(
                 f"Skipping `{market.question}` because it seems to not be predictable."
             )
             continue
 
-        if market.description and not is_predictable_without_description(
+        # We don't check this for Polymarket because, if is_predictable is false, we rephrase the question.
```
kongzii, line 156: It seems to me form code above, that Manifold questions will get rephrased as well.

I think that will happen quite often. So maybe here something like `if was_rephrased and market.description and not ...`, instead of checking for manifold?

prediction_market_agent/agents/replicate_to_omen_agent/llm.py
```diff
@@ -0,0 +1,44 @@
...
+    llm = ChatOpenAI(
+        model_name=engine,
+        temperature=LLM_SUPER_LOW_TEMPERATURE,
+        openai_api_key=APIKeys().openai_api_key,
+    )
+
+    prompt = ChatPromptTemplate.from_template(template=REPHRASE_PROMPT)
+    messages = prompt.format_messages(
+        question=question,
+        current_date=str(utcnow()),
```
kongzii, line 34: I have a new worry here, we don't know if gpt-4.1-mini is deterministic with low temperature + this time is always changing, so the rephrased question will be a bit different each time.

Because of that, the current mechanism to detect already replicated questions will fail, and we will end up with replicating the same question over and over again, just with a bit different wording.

Did you check if this is happening or not?

If it's actually happening, you can check out [this experiment from last year](https://github.com/gnosis/prediction-market-agent/pull/438), to find out how to make LLM deterministic. Then you can instead of `utcnow()` pass creation date of the market here. 

And then hopefully the rephrased questions will be the same.

prediction_market_agent/agents/replicate_to_omen_agent/llm.py
```diff
@@ -0,0 +1,44 @@
...
+
+[QUESTION]
+"{question}"
+
+[PROMPT FOR QUESTION VALIDITY]
+{invalid_question_prompt}
+"""
+
+
+def rephrase(question: str, engine: str = "gpt-4.1-mini-2025-04-14") -> str:
```
TS9001, line 24: Why 4.1 mini? Is it enough for difficult questions?

prediction_market_agent/agents/replicate_to_omen_agent/deploy.py
```diff
@@ -90,21 +98,40 @@ def replicate(self, settings: ReplicateSettings) -> None:
...
             )
 
+            if replicate_config.collateral_token:
+                collateral_token_address = Web3.to_checksum_address(
+                    replicate_config.collateral_token
+                )
+                # make sure it's ERC20
+                try:
+                    ContractERC20OnGnosisChain(
+                        address=collateral_token_address
```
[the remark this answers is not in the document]
gabrielfior: thanks, implemented

prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
gabrielfior: This was an old version - no more reordering happening now.

prediction_market_agent/agents/replicate_to_omen_agent/llm.py
gabrielfior: Please check https://github.com/gnosis/prediction-market-agent/pull/912/commits/53a87eba5fbad79dafc31d66725373cdb551871c#diff-7706eb1de513aca7a2f2070c1c3dfa6488a7aaef052cff121b96aa7292da3047R70-R82

tl;dr there is a new table in the DB that stores the replicated markets, since there is no way to tell (using the subgraph) which parent market used the child market. I thought this was easier to implement than trying to make the LLM deterministic.

prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
gabrielfior: Please check https://github.com/gnosis/prediction-market-agent/pull/912/commits/53a87eba5fbad79dafc31d66725373cdb551871c#diff-7706eb1de513aca7a2f2070c1c3dfa6488a7aaef052cff121b96aa7292da3047R180-R186

Now I'm skipping this for the case when it `was_rephrased`, because I run a check previously making sure it's predictable, so this check didn't seem necessary.

prediction_market_agent/agents/replicate_to_omen_agent/llm.py
gabrielfior: It's good enough, it rephrases things and clarifies questions enough so as to make it valid.

prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
```diff
@@ -132,23 +156,46 @@ def omen_replicate_from_tx(
...
+        # We don't check this for the was_rephrased case because the question was already considered predictable.
+        if (
+            not was_rephrased
+            and market.description
+            and not is_predictable_without_description(
+                market.question, market.description
+            )
         ):
+            # We try rephrasing the question to combine elements of the description into the question.
+            new_question = rephrase(market.question + market.description)
```
gabrielfior, line 188: Note that I added this `rephrase` because, out of ~ 200 Polymarket questions, only one was being replicated.
Example:

old_polymarket_question - Bitcoin above $118,000 on July 18?
replicated_question_after_rephrase - Will the Binance BTC/USDT 1-minute candle close above $118,000 at 12:00 ET on July 18, 2025?

prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
```diff
@@ -132,23 +156,46 @@ def omen_replicate_from_tx(
...
 
         if not is_predictable_binary(market.question):
             logger.info(
                 f"Skipping `{market.question}` because it seems to not be predictable."
             )
             continue
 
-        if market.description and not is_predictable_without_description(
-            market.question, market.description
+        # We don't check this for the was_rephrased case because the question was already considered predictable.
```
kongzii, line 179: After the rephrasing above, the question is not invalid, but it doesn't mean it's "predictable without description".

Seems to be this should be checked for even if it was rephrased.

gabrielfior: makes sense - now even after rephrasing, we still run `is_predictable_without_description`

prediction_market_agent/db/replicated_markets_table_handler.py
```diff
@@ -0,0 +1,27 @@
...
+        self.sql_handler = SQLHandler(
+            model=ReplicatedMarket, sqlalchemy_db_url=sqlalchemy_db_url
+        )
+
+    def save_replicated_markets(self, markets: list[ReplicatedMarket]) -> None:
+        """Save item to storage."""
+        self.sql_handler.save_multiple(markets)
+
+    def get_replicated_markets_from_market(
+        self, parent_market_type: str
```
kongzii, line 21:
```suggestion
        self, original_market_type: str
```

As agreed in the other comment

prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
```diff
@@ -53,22 +55,21 @@ def omen_replicate_from_tx(
...
-    initial_funds: USD,
+    initial_funds: USD | CollateralToken,
     collateral_token_address: ChecksumAddress,
     close_time_before: DatetimeUTC | None = None,
     close_time_after: DatetimeUTC | None = None,
     auto_deposit: bool = False,
     test: bool = False,
 ) -> list[ChecksumAddress]:
-    existing_markets = OmenSubgraphHandler().get_omen_markets(limit=None)
+    existing_markets = OmenSubgraphHandler().get_omen_markets(limit=10)
```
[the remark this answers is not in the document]
gabrielfior: `limit=None` set on commit https://github.com/gnosis/prediction-market-agent/pull/912/commits/faca60ef0cc4f0d41f76cb561faed3346c28f3d7

prediction_market_agent/agents/replicate_to_omen_agent/omen_replicate.py
```diff
@@ -84,6 +102,7 @@ def omen_replicate_from_tx(
         for m in markets_sorted
         if close_time_before is None
         or (m.close_time is not None and m.close_time <= close_time_before)
+        or m.question in excluded_questions
```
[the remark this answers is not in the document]
gabrielfior: my mistake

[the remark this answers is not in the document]
gabrielfior: outdated

PR #926, opened 2025-07-16
FSCA agent base created

TS9001:


prediction_market_agent/agents/full_set_collective_arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,344 @@
...
+        For over-estimated probabilities (sum > 1), we:
+        1. Mint complete sets (cost: quantity * 1.0)
+        2. Sell all outcome tokens (revenue: quantity * sum(probabilities) - slippage)
+        3. Profit = revenue - cost
+        
+        Returns optimal quantity where marginal profit = marginal cost increase
+        """
+        quantity = 0
+        
+        while True:
```
gabrielfior, line 186: Maybe I'm mistaken, but Omen has a function `calcBuyAmount` that lets you calculate how many `outcome tokens` you will get in exchange for collateral. So this way you can directly calculate the price impact (without approximation).

Something similar could be achieved for Seer as well, and would work as follows:
- You call the `exactInputSingle` method from `SwaprRouterContract`, but instead of executing it with `web3.eth.sendTransaction`, you [call](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.call) `web3.eth.call`, so you simply get the returned `output token amounts` back.

prediction_market_agent/agents/full_set_collective_arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,421 @@
...
+    ) -> t.Sequence[SeerAgentMarket]:
+        # Fetch all markets to choose from
+        available_markets = SeerAgentMarket.get_markets(
+            limit=5,
+            # sort_by=self.get_markets_sort_by,
+            sort_by=SortBy.HIGHEST_LIQUIDITY,
+            filter_by=self.get_markets_filter_by,
+            created_after=self.trade_on_markets_created_after,
+        )
+        return available_markets
```
gabrielfior, line 81: Does this need to be overriden? I would've expected the [logic](https://github.com/gnosis/prediction-market-agent-tooling/blob/6897bb44a8a8fffe35c564302fc6b0e9bd0afae0/prediction_market_agent_tooling/markets/seer/seer.py#L449) from `SeerAgentMarket` to be enough

prediction_market_agent/agents/full_set_collective_arbitrage_agent/deploy.py
```diff
@@ -0,0 +1,421 @@
...
+
+    def process_market(
```
gabrielfior, line 83: You can override the method `before_process_market` and pack the logic there - this way, I believe you can avoid placing the trades manually.

PR #932, opened 2025-07-24
Bert Trading Agent (binary classifier)

kongzii:
**Hypothesis:** Because binary markets are 50/50, a normal classifier could be trained to be just a little correct (although it couldn't predict the future in practice, ofc), and be profitable in the end.

**Conclusion:** Although it works, and the trained classifier is indeed profitable, it's not more profitable than just saying always "NO", because of the big skew in the data, as shown below. After training, the model has decent precision/recall on the positive class, but is still behind a fixed NO classifier; that's why I wouldn't deploy it for now, but keep the code in a closed PR for future reference. 

**Next steps:** I will deploy a statistical fixed agent based on the discovery here. And, if the skew ever goes lower (currently ~65% of markets are resolved as NO), then this trained classifier could be more profitable than a fixed agent. But that needs to be verified on actual data once it happens.

**Details for those who are interested:**

I fetch all resolved markets from 1.10.2024 (after OFVChallenger was deployed), to have a reasonable resolution accuracy in the data. The amount of markets is quite stable in time:

image

I split them into train/val/test based on the timeline, last 6 weeks for testing, 6 weeks before for validation, and the rest for training.

The skew is visible:

```
Train: 4198, Val: 870, Test: 941
Label proportions: train={0: 0.6245831348261077, 1: 0.37541686517389233}, val={0: 0.6689655172413793, 1: 0.3310344827586207}, test={0: 0.6748140276301806, 1: 0.32518597236981933}
```

To deal with the skew, I tried:

1. Ignore it
2. Oversample minority class
3. Undersample majority class
4. Synthetise data with LLM

(2) is in the code as it worked best.

I use `distilbert-base-uncased` for the model, I tried several others from HuggingFace, but this worked best.

I set `max_length` in tokenisation to `42`, because this is how the distribution looks like:

image

For hyperparameters, see the code, but it standard distilbert finetuning + early stopping + learning rate schedule.

For the loss function, BCE is used in the code. I also tried to optimize directly for profit (loss being the negative mean expected value from the prediction), but it didn't give better results.

Interesting result from training is that the predicted probability is kind of all over the place, so I tried to calibrate it:

image

And then, I tried to optimize the threshold value based on the validation set, unfortunatelly that basically convergs to always saying "NO":

image

After the training, metrics on training set are all pefrect, so model is cabable to just remember answers on the training set (don't be fooled by the `test` name in the screenshot, the name is bug, metrics aren't):

Screenshot by Dropbox Capture

Unfortunately, on the validation and testing set, "NO-Agent" is better than BERT 🤷:

Screenshot by Dropbox Capture

I also tried to train the model against the market's predicted probability (instead of the final resolution), but that also didn't work well.


kongzii:
Closing just to keep the code somewhere.

PR #940, opened 2025-08-05
Bump pmat 0.67.4

kongzii:


gabrielfior (approved):

kongzii:
Will merge to get the logs in, these failing test now fail even on PRs from last week. So unrelated to this.

PR #957, opened 2025-08-18
Tested betting differences

TS9001:


prediction_market_agent/development_tools/prophet_agent_tester.py
```diff
@@ -201,10 +250,8 @@ def test_prophet_agent(
                     prediction=prediction,
                     trades=trades,
                     market_resolution=item["market_resolution"],
-                    received_outcome_tokens=float(received_outcome_tokens.as_token)
-                    if received_outcome_tokens
-                    else None,
-                    profit_outcome_token=float(profit_outcome_token),
+                    received_outcome_tokens=None,
```
kongzii, line 253: Why set received_outcome_tokens to None instead of providing them, if available, as before?

TS9001: Fixed, thank you for the catch

Issue #946, opened 2025-08-06
Arbitrage Agent is in loss


kongzii:
Which is weird, as he should be arbitrageur to profitability, right? https://presagio.pages.dev/profile?address=0x2fc96c4e7818dbdc3d72a463f47f0e1ceea0a2d0

PR #960, opened 2025-08-19
Fixed arbitrage agent loosing money

TS9001:
Fixes: https://github.com/gnosis/prediction-market-agent/issues/946

### DIFF
--- a/prediction_market_agent/agents/arbitrage_agent/deploy.py
+++ b/prediction_market_agent/agents/arbitrage_agent/deploy.py
@@ -1,6 +1,7 @@
 import typing as t
 from datetime import timedelta
 
+from cachetools import TTLCache
 from langchain_core.output_parsers import PydanticOutputParser
 from langchain_core.prompts import PromptTemplate
 from langchain_core.runnables import RunnableSerializable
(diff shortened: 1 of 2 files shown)


prediction_market_agent/agents/arbitrage_agent/deploy.py
```diff
@@ -194,5 +235,26 @@ def build_trades(
...
+                    )
 
         return trades
+
+    def execute_trades(
+        self, market: AgentMarket, trades: list[Trade]
+    ) -> list[PlacedTrade]:
+        placed_trades: list[PlacedTrade] = []
+        for trade in trades:
+            if isinstance(trade, MarketTrade):
```
gabrielfior, line 253: I don't follow this `isistance` check - how would it be possible that trades of another type are produced? See methods below for reference

```
trades_for_pair = self.build_trades_for_correlated_markets(pair)

@observe()
    def build_trades_for_correlated_markets(
        self, pair: CorrelatedMarketPair
    ) -> list[MarketTrade]:
```

TS9001: You are right, unfortunately mypy thinks otherwise... :facepalm: :man_facepalming: :woman_facepalming:

TS9001: Yes it is but it is also generally bad practice to change general class because of single effect - in this case it is probably good idea but I would rather refactor whole method to not to accept market and use one provided. This would fix the issue but affect lots of agents - would you be ok to do it in specific issue and link this conversation to it?

PR #961, opened 2025-08-20
Bump PMAT 0.69.1

kongzii:


prediction_market_agent/agents/gptr_agent/deploy.py
```diff
@@ -6,7 +6,7 @@
 from prediction_market_agent_tooling.deploy.agent import DeployableTraderAgent
 from prediction_market_agent_tooling.deploy.betting_strategy import (
     BettingStrategy,
-    BinaryKellyBettingStrategy,
+    FullBinaryKellyBettingStrategy,
```
kongzii, line 9: Full was the default go-to in the previous BinaryKellyBettingStrategy, that's why I replaced it with this one.

In next PR, I will reconfigure the strategies based on the optimization script.

PR #964, opened 2025-08-22
Maybe fix ofvchallenger, at least double bound error

kongzii:


TS9001 (approved):

gabrielfior:
If the subgraph is unreliable, one can also fetch the information in other ways, e.g. onchain, deploy a new subgraph, etc.
Are you sure that simply retrying solves it?

PR #988, opened 2026-04-09
Lower n bet per run on agents

kongzii:


prediction_market_agent/agents/advanced_agent/deploy.py
```diff
@@ -19,7 +19,7 @@ class AdvancedAgent(DeployableTraderAgent):
     Use as a baseline for comparing with other agents.
     """
 
-    bet_on_n_markets_per_run = 4
+    bet_on_n_markets_per_run = 1
```
kongzii, line 22: Agents ate all the Tavily credits in a day. I guess we had the biggest plan ($500 monthly) back then.

I upgraded the plan to $220/mo (38k credits).

But I think we can also back down on the number of bets and increase it again bit by bit based on our progress?

TS9001 (approved):

